PR #280, opened 2021-01-26
Limit deposits using the minted LP token amount instead of balance

### LINKED ISSUE #265: Read minted amounts from the LP token
Individual deposits are limited by comparing `LPToken.mintedAmount(account)` to `allowlist.getPoolAccountLimit(poolAddress)`

We should update the frontend to behave the same as the limits in the smart contract.

https://etherscan.io/token/0xc28df698475dec994be00c9c9d8658a548e6304f#readContract

alphastorm:
Closes #265


hammeiam (approved):
lgtm!

PR #292, opened 2021-01-30
Update to create-react-app v4

alphastorm:
Close #134
Close #175


sandrasong (approved):

src/hooks/useWithdrawFormState.ts
```diff
@@ -216,8 +216,8 @@ export default function useWithdrawFormState(
       }
       setFormState((prevState) => ({
         ...prevState,
-        error: null,
         ...nextState,
+        error: null,
```
hammeiam, line 220: We do want to clear out the previous state's error, but we don't want to overwrite the nextState's error - was this a suggested change?

src/hooks/useWithdrawFormState.ts
```diff
@@ -285,8 +285,8 @@ export default function useWithdrawFormState(
         }
         const finalState = {
           ...prevState,
-          error: null,
           ...nextState,
+          error: null,
```
hammeiam, line 289: ditto

hammeiam (approved):
One question about overriding a property in an object, lgtm otherwise - approving to unblock

PR #308, opened 2021-02-04
lpToken amount in Deposit Review, new TransactionInterface, safer TokenInput

hammeiam:
- Added lpToken price USD to deposit review page
- created Transaction interfaces, refactored deposits to employ it. @alphastorm I'd love your input on a cleaner way to account for having either `item` or `items` while still having aggregated (`totalValueUSD`, `totalAmount`) in a DRY way
- Made TokenInput component safer - now it's impossible to enter an input of higher precision than the target token

@sandrasong wrt modifying the design of the review modal, my opinion is that just having the rates (`1 BTC = $40k`) isn't really helpful or actionable in the review screen. I think maybe summing the number of tokens being deposited to more easily compare to the number of lpTokens being received could be helpful. And somehow doing the same thing for USD. Like we discussed when refactoring the `historicalData` component, having a button to toggle between token and USD values might be nice and make comparison easy. What do you think?

Screen Shot 2021-02-04 at 12 36 21 AM


sandrasong:
@hammeiam  I agree just having the rates is not helpful. The USD calculation is hard with so many decimals. Users would hate to do it. 

Having a toggle button may be a bit too much for a modal. Model's interaction is expected to be very simple, something like quickly check the information and click action button. What do you think of the below display?

review_pool_LP rate

hammeiam:
> @hammeiam I agree just having the rates is not helpful. The USD calculation is hard with so many decimals. Users would hate to do it.
> 
> Having a toggle button may be a bit too much for a modal. Model's interaction is expected to be very simple, something like quickly check the information and click action button. What do you think of the below display? (Changed the outline to be grey to standout from github background. It's white in dApp.)
> 
> review_pool_LP rate (1)

@sandrasong honestly I'm not sure how useful the rates information is at all. I don't ever see someone changing their deposit behavior due to the current USD price of an asset. I think that as a user, the most impactful information for me is:
- how many of each token I'm depositing
- how many total tokens I'm depositing
- how many lpTokens I'm getting in return 
- if there is a high price impact
- my advanced options

How would you feel about me adding a "sum", but removing all of the rates? That should have the net effect of removing information from the modal.

hammeiam:
> > > honestly I'm not sure how useful the rates information is at all.
> 
> True that I don't see other products have this information. So probably the `high price impact` warning is enough.
> @hammeiam

Sounds good, thanks! I'll go with my proposed plan for now then. 
If we wanted to make modals wider, we could adopt something like Debank has
Screen Shot 2021-02-08 at 1 11 24 PM

sandrasong (approved):

sandrasong (approved):

alphastorm (approved):

PR #310, opened 2021-02-04
Allow quoting without balance

### LINKED ISSUE #274: Quote trades even if the user doesn't have the balance to perform them
Currently if you type a value into the `From` field without actually holding the token balance the user is given an `Insufficient balance` error.  We should also quote the expected amount of received tokens in the `To` field.

hammeiam:
Closes #274


bobtista (approved):
LGTM

alphastorm (approved):

PR #315, opened 2021-02-09
Add custom transaction deadlines for deposits and withdrawals

TheGreatGildo:
added deadline field to deposit and withdraw page
added deadline information in the review deposit and review withdraw modals
changed the ordering of the advanced options to be consistent throughout the site
renamed "Transaction Deadline" to "Txn Deadline" -- fixes a lot of ugly formatting issues when there is a line break
Screenshot 2021-02-09 13:53:27
Screenshot 2021-02-09 13:53:02


.env
```diff
@@ -1,3 +1,3 @@
 REACT_APP_CHAIN_ID="1"
 REACT_APP_GOOGLE_ANALYTICS_MEASUREMENT_ID=""
-REACT_APP_NETWORK_URL=""
+REACT_APP_NETWORK_URL="https://eth-mainnet.alchemyapi.io/v2/qNhZ_-GaDehhI-HSlG_dff9nxS6nuXGg"
```
alphastorm, line 3: this should be set in an `.env.local` file so it isn't committed to git.  in general, you never want to commit secrets or keys to repositories.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
alphastorm, line 1: editor specific functionality also shouldn't be committed, you can add the `.vscode/` directory to the `.gitignore`

alphastorm (commented):
lgtm after addressing the two comments

.env
TheGreatGildo: fixed

.vscode/settings.json
TheGreatGildo: fixed

src/hooks/useApproveAndDeposit.ts
```diff
@@ -142,7 +143,9 @@ export function useApproveAndDeposit(
       const spendTransaction = await swapContract.addLiquidity(
         POOL_TOKENS.map(({ symbol }) => state.tokenFormState[symbol].valueSafe),
         minToMint,
-        Math.round(new Date().getTime() / 1000 + 60 * 10),
+        Math.round(
+          new Date().getTime() / 1000 + 60 * state.transactionDeadline,
```
hammeiam, line 147: is `state.transactionDeadline` always guaranteed to be populated at this point?

hammeiam (approved):
generally lgtm, let's just confirm that `state.transactionDeadline` is always populated with something by the time we try to do math on it.

sandrasong (approved):

src/hooks/useApproveAndDeposit.ts
TheGreatGildo: double checked and the logic for this is handled in the deadline field and works across all deadline fields regardless the page

found in DeadlineField.tsx
`onChange={(e): void => {
            const deadlineNumber =
              e.target.value === "" || parseFloat(e.target.value) <= 0
                ? Deadlines.Ten // if left blank, deadline defaults to 10
                : parseFloat(e.target.value)
            dispatch(updateTransactionDeadline(deadlineNumber))
          }}`

PR #318, opened 2021-02-12
Hide Historic Data

hammeiam:
Hide this tab until we're able to build a more correct solution. Users have reported issues with the validity of the data we were showing, so don't show them invalid data for now.


alphastorm (approved):
lgtm after resolving merge conflicts

PR #348, opened 2021-02-24
use Zero from ethers constants

bobtista:
We were using a mix of bignumber.from(0) and Zero from ethers constants. 
This PR makes things more consistent, and I think neater.


hammeiam (approved):
thanks!

PR #350, opened 2021-02-25
Correct deadline loading logic

hammeiam:
See [slack](https://saddle-finance.slack.com/archives/C017XJU4GSJ/p1614222819021900) for issue details.

We were incorrectly merging static `initialState` with state pulled from `localStorage` while hydrating redux. As a result, some users did not have a `transactionDeadline`, which consequently caused their transactions to fail (before even reaching the contract, so no gas was spent). 

This PR does 3 things in order of importance:
1) Corrects how `initialState` is merged into localstorage state in redux
2) Adjusts the implementation of `deadlines` to match that of `slippage` and `gas` (eg using `selected` and `custom` states) which also fixes a minor bug (type 10 or 30 into the custom deadline field in prod to see)
3) Moves Deadlines, gas, and slippage from props to being accessed from the redux store in all `useApproveAnd[Swap|Deposit|Withdraw]` functions

### DIFF
--- a/src/components/DeadlineField.tsx
+++ b/src/components/DeadlineField.tsx
@@ -55,13 +55,6 @@ export default function DeadlineField(): ReactElement {
           onClick={(): PayloadAction<Deadlines> =>
             dispatch(updateTransactionDeadlineSelected(Deadlines.Custom))
           }
-          // onChange={(e): void => {
-          //   const deadlineNumber =
-          //     e.target.value === "" || parseFloat(e.target.value) <= 0
-          //       ? Deadlines.Ten // if left blank, deadline defaults to 10
-          //       : parseFloat(e.target.value)
-          //   dispatch(updateTransactionDeadline(deadlineNumber))
-          // }}
           value={transactionDeadlineCustom}
           onChange={(e: React.ChangeEvent<HTMLInputElement>): void => {
             const value = e.target.value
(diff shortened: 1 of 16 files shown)


src/state/index.ts
```diff
@@ -15,7 +18,7 @@ const store = configureStore({
     ...getDefaultMiddleware({ thunk: false }),
     save({ states: PERSISTED_KEYS }),
   ],
-  preloadedState: load({ states: PERSISTED_KEYS }),
+  preloadedState: merge({}, { user: userInitialState }, stateFromStorage),
```
hammeiam, line 21: IMPORTANT BIT HERE

src/components/DeadlineField.tsx
```diff
@@ -1,64 +1,81 @@
...
-                ? Deadlines.Ten // if left blank, deadline defaults to 10
-                : parseFloat(e.target.value)
-            dispatch(updateTransactionDeadline(deadlineNumber))
+          // onChange={(e): void => {
+          //   const deadlineNumber =
+          //     e.target.value === "" || parseFloat(e.target.value) <= 0
+          //       ? Deadlines.Ten // if left blank, deadline defaults to 10
+          //       : parseFloat(e.target.value)
+          //   dispatch(updateTransactionDeadline(deadlineNumber))
+          // }}
```
alphastorm, line 64: nit: remove

alphastorm (approved):

PR #372, opened 2021-03-10
Make existing pages generic w/ poolName prop

hammeiam:
Remove hardcoded poolnames from top level pages, replace with `poolName` prop. Check `App.tsx` for how this is used.
Tested and working locally on a branch of the contract repo before some recent changes to the local environment.

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -64,6 +64,13 @@ export const MERKLETREE_DATA: { [chainId in ChainId]: string } = {
   [ChainId.HARDHAT]: "hardhat.json",
 }
 
+export const STABLECOIN_SWAP_TOKEN_CONTRACT_ADDRESSES: {
+  [chainId in ChainId]: string
+} = {
+  [ChainId.MAINNET]: "",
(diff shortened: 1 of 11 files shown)


src/constants/index.ts
```diff
@@ -76,7 +83,16 @@ export const BTC_SWAP_TOKEN = new Token(
...
   "blpt",
-  "Saddle BTC Pool LP Token",
+  "Saddle BTC LP Token",
+  saddleLogo,
+)
+
+export const STABLECOIN_SWAP_TOKEN = new Token(
+  STABLECOIN_SWAP_TOKEN_CONTRACT_ADDRESSES,
+  18,
+  "SLPT",
```
hammeiam, line 93: I'm curious about the correctness of these names / symbols

penandlim: For reference, on-chain USD LP token will follow the btc lp token's convention
`Saddle DAI/USDC/USDT` for the name
`saddleUSD` for symbol

alphastorm (approved):

src/constants/index.ts
hammeiam: Thanks, will update!

penandlim (approved):

PR #373, opened 2021-03-11
USDPool + new swap interface

hammeiam:
This PR is branched off of the new swap interface PR; thus, all work for this PR is contained in `07d3bdd` and after. The other PR will be landed first (when we're ready for multipool) then this will be landed. 

- Add new routes for deposit/withdraw usd/btc
- Create `Pools` component to serve as a pool index view
- minor changes to usePoolData

DO NOT LAND until usd pool is deployed, its address is entered in THIS pr, and the token composition is corrected in this pr.


penandlim:
Updated the contract addresses to reflect the changes from https://github.com/saddle-finance/saddle-contract/pull/323

penandlim (approved):

PR #400, opened 2021-03-29
Zero values instead of contract data when no user wallet connected

hammeiam:
Since we're unable to interact with contracts when the user is not connected to a wallet, just show 0 values for everything to give a less broken experience. Additionally, display the "connect wallet" modal on every page until the user connects a wallet. 

Screen Shot 2021-03-29 at 9 14 21 AM
Screen Shot 2021-03-29 at 9 14 57 AM
Screen Shot 2021-03-29 at 9 14 39 AM

### DIFF
--- a/src/components/PoolOverview.tsx
+++ b/src/components/PoolOverview.tsx
@@ -13,7 +13,7 @@ import { useTranslation } from "react-i18next"
 
 interface Props {
   to: string
-  poolData: PoolDataType | null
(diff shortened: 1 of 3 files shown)


alphastorm (approved):
this is good for now, but i don't think we should have the modal pop up on every page once we're using the default provider and data is populated w/o the wallet being connected

PR #405, opened 2021-04-01
Show keep staking

hammeiam:
Query the Keep/Saddle LP Staking contract for amount of tokens staked in KEEP. Present that figure to the user in the MyShare component. I'd love some someone who's staked to try this out. 

Screen Shot 2021-03-31 at 5 09 07 PM


src/constants/abis/keepLPRewards.json
```diff
@@ -0,0 +1,13 @@
+[
```
alphastorm, line 1: nit: i think you might be able to use `src/constants/abis/erc20.json` instead of adding a new ABI

src/components/MyShareCard.tsx
```diff
@@ -46,7 +50,12 @@ function MyShareCard({ data }: Props): ReactElement | null {
         </div>
         <div className="infoItem">
           <span className="label bold">{`${t("totalAmount")}: `}</span>
-          <span className="value">{formattedData.value}</span>
+          <span className="value">{formattedData.amount}</span>
+          {data.amountStakedOnKeep.gt(Zero) ? (
+            <span className="value">
+              &nbsp;({formattedData.amountStakedOnKeep} on KEEP)
```
alphastorm, line 56: nit: i think saying `staked` (maybe with a hyperlink to the rewards dashboard) is more clear than `on KEEP`, also could we add a translation for this

src/hooks/usePoolData.ts
```diff
@@ -127,6 +129,18 @@ export default function usePoolData(
           account ?? undefined,
         ) as LpTokenUnguarded
       }
+      const keepLPRewardsContract = process.env.production
```
alphastorm, line 132: nit: i'd add `&& poolName === BTC_POOL_NAME` so we only make this call for that specific pool

alphastorm (approved):
overall lgtm, but it'd be awesome if @penandlim can double check on mainnet 🤠

src/components/MyShareCard.tsx
penandlim: +1 on hyperlink to the rewards dashboard

src/constants/abis/keepLPRewards.json
penandlim: Though the erc20.json would work, for the sake of clarity, we shouldn't use ABIs that don't fully match the contract.

penandlim:
I was able to see the staked amounts on mainnet via fleek preview 👍

penandlim (approved):

PR #451, opened 2021-05-04
Fetch Pending Swaps

hammeiam:
Creates `usePendingSwapData.ts` hook which fetches existing virtual swaps from event logs on mount, and also attaches listeners for new virtual swap actions. Also adds the basic UI of showing that there are pendingSwaps.

Screen Shot 2021-05-28 at 12 06 15 PM

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -63,6 +63,10 @@ export class Token {
 }
 
 export const BLOCK_TIME = 15000
+export const DEPLOYED_BLOCK: { [chainId in ChainId]: string } = {
+  [ChainId.MAINNET]: "latest",
+  [ChainId.HARDHAT]: "0",
+}
 
 export const BRIDGE_CONTRACT_ADDRESSES: { [chainId in ChainId]: string } = {
   [ChainId.MAINNET]: "0xf5059a5D33d5853360D16C683c16e67980206f36", // TODO replace once mainnet deploy goes out
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -62,11 +62,7 @@ export class Token {
   }
 }
 
-export const BLOCK_TIME = 15000
-export const DEPLOYED_BLOCK: { [chainId in ChainId]: string } = {
-  [ChainId.MAINNET]: "latest",
-  [ChainId.HARDHAT]: "0",
-}
+export const BLOCK_TIME = 13000 // ms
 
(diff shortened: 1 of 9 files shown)


penandlim:
Looks good to me so far. 
* Dudupeing by `itemId` is preferred. Even after a virtual swap is completed the once used itemId will not be overwritten.
* Currently WIP. Since the bridge requires synthetix stack to be deployed, we can only deploy bridge in a forked network.

src/constants/index.ts
```diff
@@ -62,7 +62,7 @@ export class Token {
   }
 }
 
-export const BLOCK_TIME = 15000
+export const BLOCK_TIME = 13000 // ms
```
hammeiam, line 65: I was checking out etherscan and it looks like the average is about 13s

src/hooks/usePendingSwapData.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  // update the secondsRemaining every 15s
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setPendingSwaps((swaps) =>
+        swaps.map((swap) => {
+          let secondsRemaining = Math.floor(
+            (swap.settleableAtTimestamp.getTime() - Date.now()) / 1000,
+          )
+          secondsRemaining = secondsRemaining > 0 ? secondsRemaining : 0
```
alphastorm, line 49: nit: easier to understand
```js
secondsRemaining = Math.max(secondsRemaining, 0)
```

alphastorm (approved):

src/hooks/usePendingSwapData.ts
```diff
@@ -0,0 +1,214 @@
...
+    const settleableAtTimestamp = new Date(
+      (timestampInSeconds + parseInt(pendingSwapInfo.secsLeft.toString())) * // add event block timestamp + secsLeft
+        1000, // convert to ms
+    )
+    result = {
+      swapType,
+      settleableAtTimestamp,
+      secondsRemaining: Math.ceil(
+        (Date.now() - settleableAtTimestamp.getTime()) / 1000,
+      ),
```
penandlim, line 202: Note that `Date.now()` will always be head of `block.timestamp` (latest block's timestamp). Ideally we should be checking the remaining time against `(await ethers.provider.getBlock("latest")).timestamp` or read from the multicall contract https://github.com/makerdao/multicall/blob/master/src/Multicall.sol#L33-L35

PR #459, opened 2021-05-13
SharedStake APY

hammeiam:
- Calculate veth2 price from the ex-rate between veths & eth. 
- Add sgt pool apy in `usePoolData` hook. 
- Display new apy. 
- Nest tokenPrices under react manager so that the token price calculations have access to a web3 provider

NOTE: Our apy is slightly different from what is shown on the SGT website. I think this a bug in [their code](https://github.com/SharedStake/SharedStake-ui/blob/main/src/utils/veth2.js) where they calculate the veth2 price by calling `swap.calculateSwap(0,1)` which is eth -> weth2, when what I think they want is `(1,0)`. But somebody please double check my math on this as I am no mathlete.

Screen Shot 2021-05-11 at 11 50 15 PM

### DIFF
--- a/src/hooks/usePoolData.ts
+++ b/src/hooks/usePoolData.ts
@@ -17,6 +17,8 @@ import LPTOKEN_GUARDED_ABI from "../constants/abis/lpTokenGuarded.json"
 import LPTOKEN_UNGUARDED_ABI from "../constants/abis/lpTokenUnguarded.json"
 import { LpTokenGuarded } from "../../types/ethers-contracts/LpTokenGuarded"
 import { LpTokenUnguarded } from "../../types/ethers-contracts/LpTokenUnguarded"
+import SGT_REWARDS_ABI from "../constants/abis/sharedStakeStakingRewards.json"
+import { SharedStakeStakingRewards } from "../../types/ethers-contracts/SharedStakeStakingRewards"
 import { SimpleBalanceOf } from "../../types/ethers-contracts/SimpleBalanceOf"
 import { parseUnits } from "@ethersproject/units"
 import { useActiveWeb3React } from "."
@@ -40,7 +42,10 @@ export interface PoolDataType {
   utilization: string // TODO: calculate
   virtualPrice: BigNumber
   volume: string // TODO: calculate
-  keepApr: BigNumber
+  aprs: {
+    keep: BigNumber
+    sharedStake: BigNumber
+  }
   lpTokenPriceUSD: BigNumber
 }
 
@@ -68,7 +73,10 @@ const emptyPoolData = {
   utilization: "",
   virtualPrice: Zero,
   volume: "",
-  keepApr: Zero,
+  aprs: {
+    keep: Zero,
+    sharedStake: Zero,
+  },
   lpTokenPriceUSD: Zero,
 } as PoolDataType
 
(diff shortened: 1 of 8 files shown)


src/pages/App.tsx
```diff
@@ -19,83 +19,92 @@ import Web3ReactManager from "../components/Web3ReactManager"
 import Withdraw from "./Withdraw"
 import fetchGasPrices from "../utils/updateGasPrices"
 import fetchTokenPricesUSD from "../utils/updateTokenPrices"
+import { useActiveWeb3React } from "../hooks"
 import { useDispatch } from "react-redux"
 import usePoller from "../hooks/usePoller"
 
 export default function App(): ReactElement {
+  return (
+    <Suspense fallback={null}>
```
hammeiam, line 28: This looks like a big change but all I did was move the price polling stuff into `GasAndTokenPrices`, then nest everything underneath that new component.

src/hooks/usePoolData.ts
```diff
@@ -334,3 +355,32 @@ export default function usePoolData(
...
+  sgtPrice = 0,
+  tvlUsd: BigNumber,
+): Promise<BigNumber> {
+  // https://github.com/SharedStake/SharedStake-ui/blob/main/src/components/Earn/geyser.vue#L336
+  if (rewardsContract == null || tvlUsd.eq(Zero)) return Zero
+  const now = BigNumber.from(Math.floor(Date.now() / 1000))
+  const [until, rewardsDuration, sgtRewardsPerPeriod] = await Promise.all([
+    rewardsContract.periodFinish(), // 1e0 timestamp in seconds
+    rewardsContract.rewardsDuration(), // 1e0 seconds
+    rewardsContract.getRewardForDuration(), // 1e18
```
alphastorm, line 370: nit: this would be a great place to use multicall if possible so it's only 1 request

src/hooks/usePoolData.ts
```diff
@@ -334,3 +355,32 @@ export default function usePoolData(
 
   return poolData
 }
+
+async function getSgtApy(
```
alphastorm, line 359: nit: this should be APR (also applies to references in the PR title/description)

src/types/ethcall.d.ts
```diff
@@ -17,7 +17,37 @@ export type MulticallContract<ContractType> = {
...
 export class MulticallProvider extends Provider {
-  // TODO: this only support return values of the same type rather than mixed
-  // TS can't infer tuple types :(
-  all<O>(calls: MulticallCall<I, O>[], overrides: CallOverrides): Promise<O[]>
+  /**
+   *  TODO: TS can't infer tuple types
+   *  so this huge type supports either arrays of calls with the same return type (eg BigNumber[])
+   *  Or a tuple of up to 4 ditinct calls *as const* (meaning you can use map or push)
+   *  eg `const calls = [getNum(), getStr(), getBN()] as const; await p.all(calls)`
+   *  Want do build a larger tuple of calls? Just add more to this definition.
```
alphastorm, line 25: nit: s/Want do/Want to/

src/types/ethcall.d.ts
```diff
@@ -17,7 +17,37 @@ export type MulticallContract<ContractType> = {
 } &
   Contract
 export class MulticallProvider extends Provider {
-  // TODO: this only support return values of the same type rather than mixed
-  // TS can't infer tuple types :(
-  all<O>(calls: MulticallCall<I, O>[], overrides: CallOverrides): Promise<O[]>
+  /**
+   *  TODO: TS can't infer tuple types
+   *  so this huge type supports either arrays of calls with the same return type (eg BigNumber[])
+   *  Or a tuple of up to 4 ditinct calls *as const* (meaning you can use map or push)
```
alphastorm, line 23: nit: s/ditinct/distinct

alphastorm (approved):
couple nits about comments, but lgtm otherwise 🚀

penandlim (approved):
Looks good to me. 
You are right they should be calling `calculateSwap(1,0)` for calculating `# of WETH per 1 vETH2`.

PR #473, opened 2021-05-25
Navigation update

sandrasong:
**Updated screenshot with right section right aligned**
Screen Shot 2021-05-26 at 10 50 57 PM

Screen Shot 2021-05-26 at 10 50 50 PM

Demonstration of deposit and withdraw:

Part 1

Part 2


src/components/PoolOverview.tsx
```diff
@@ -44,56 +44,76 @@ function PoolOverview({
     }),
   }
 
+  const hasShare = parseFloat(formattedData.userBalanceUSD) ? true : false
```
hammeiam, line 47: I wouldn't expect parseFloat to work with commified values, plus we prefer not to parse formatted values when we can just use the raw value
`const hasShare = !!userShareData?.usdBalance.gt("0")`

src/styles/theme.scss
```diff
@@ -10,6 +10,11 @@ $white: #ffffff;
 $black: #000;
 $error: #fb5a5a;
 $black80: #333333;
+$lightBlue: #e8e5fb;
+$darkPurple: #110040;
+$blue: #2a4bff;
+$indigo20T: #{$primary-blue}33;
```
hammeiam, line 16: I can't wait to come up with a standard naming convention for these 😅

src/styles/theme.scss
```diff
@@ -10,6 +10,11 @@ $white: #ffffff;
 $black: #000;
 $error: #fb5a5a;
 $black80: #333333;
+$lightBlue: #e8e5fb;
```
hammeiam, line 13: we seem to already use `this-format` for color names rather than `thisFormat`, can we keep it consistent?

src/styles/theme.scss
```diff
@@ -10,6 +10,11 @@ $white: #ffffff;
...
+$darkPurple: #110040;
+$blue: #2a4bff;
+$indigo20T: #{$primary-blue}33;
+$blue40T: #{$blue}66;
```
hammeiam, line 17: What is the extra 66? Does hex support alpha now? I'm guessing the `T` stands for transparent?

src/components/PoolOverview.scss
```diff
@@ -56,32 +47,41 @@ div.poolOverview {
...
       margin-right: 4px;
-      vertical-align: text-top;
+      vertical-align: text-bottom;
     }
   }
 
   div.right {
-    & > div {
-      margin-right: 40px;
+    display: grid;
```
hammeiam, line 55: I think we also want a `justify-content: flex-end;` here as well to pull the content right

public/locales/en.json
```diff
@@ -82,9 +82,8 @@
   "riskIntro2": "reading the code",
   "riskIntro3": "and understanding the risks involved with being a Liquidity Provider (LP) and/or using the AMM to trade pegged value crypto assets.",
   "riskLossOfPeg": "If one of the assets in a pool significantly depegs, it will effectively mean that pool liquidity providers will be left holding only that asset.",
-  "route": "Route",
-  "selectAPool": "Select a Pool",
   "shareOfPool": "Share of Pool",
+  "route": "Route",
```
alphastorm, line 86: nit: the keys are alphabetized, this should be above L85 like it was before

src/components/PoolOverview.tsx
```diff
@@ -44,56 +44,76 @@ function PoolOverview({
...
           </div>
+        )}
+        <div className="tokens">
+          <span style={{ marginRight: "8px" }}>[</span>
+          {formattedData.tokens.map((token) => (
+            <div className="token" key={token.symbol}>
+              <img alt="icon" src={token.icon} />
+              <span>{token.name}</span>
+            </div>
+          ))}
```
alphastorm, line 66: nit: can destructure here to save a few characters
```js
          {formattedData.tokens.map(({symbol, icon, name}) => (
            <div className="token" key={symbol}>
              <img alt="icon" src={icon} />
              <span>{name}</span>
            </div>
          ))}
```
ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

src/styles/theme.scss
sandrasong: Yep, that's the alpha with hex. T is for transparent.

src/components/PoolOverview.scss
sandrasong: Just did it. Forgot about the right align part.

penandlim (approved):
Tested via fleek. LGTM!

alphastorm:
lgtm, but something is breaking the cypress CI coverage.  @sandrasong you will need to run `npm run cy:coverage` locally to debug what is breaking to fix it so the build passes.  @hammeiam or i can help if you run into problems with this, let us know 😁

alphastorm (approved):

PR #478, opened 2021-06-01
Update ethcall library to support Multicall2

penandlim:
Updated ethcall library to support calling `Multicall2.tryAggregate`

`ethcallProvider.tryAll()` will ignore any reverted reads and return null values in place of the failed read call's return values.

If the changes https://github.com/Destiner/ethcall/pull/5 are approved merged in, we should switch back to the npm version.


sandrasong (approved):

penandlim:
Please do not merge this until ethcall package has been updated

PR #479, opened 2021-06-03
Display native pool APY and 24h volume on the pools page

bobtista:
Pull in swap stats from json API, add to redux state, thread through to UI. Add new short format option for bigNumbers. 
Note: We'll need to add both the alUSD pool and the 4pool to the graph for this to be useful across the board

Screen Shot 2021-06-28 at 7 39 52 PM

### DIFF
--- a/src/components/PoolOverview.tsx
+++ b/src/components/PoolOverview.tsx
@@ -3,7 +3,11 @@ import "./PoolOverview.scss"
 import { POOLS_MAP, PoolTypes, TOKENS_MAP } from "../constants"
 import { Partners, PoolDataType, UserShareType } from "../hooks/usePoolData"
 import React, { ReactElement } from "react"
-import { formatBNToPercentString, formatBNToString } from "../utils"
+import {
+  formatBNToPercentString,
+  formatBNToShortString,
+  formatBNToString,
+} from "../utils"
 
 import Button from "./Button"
 import { Link } from "react-router-dom"
@@ -28,7 +32,7 @@ function PoolOverview({
   const formattedData = {
     name: poolData.name,
     reserve: poolData.reserve
-      ? commify(formatBNToString(poolData.reserve, 18, 0))
+      ? formatBNToShortString(poolData.reserve, 18)
       : "-",
     aprs: Object.keys(poolData.aprs).reduce((acc, key) => {
       const apr = poolData.aprs[key as Partners]?.apr
@@ -41,10 +45,10 @@ function PoolOverview({
     }, {} as Partial<Record<Partners, string>>),
     apy: poolData.apy ? `${formatBNToPercentString(poolData.apy, 18, 2)}` : "-",
     volume: poolData.volume
-      ? `$${commify(formatBNToString(poolData.volume, 18, 2))}`
+      ? `$${formatBNToShortString(poolData.volume, 18)}`
       : "-",
     userBalanceUSD: commify(
-      formatBNToString(userShareData?.usdBalance || Zero, 18, 2),
+      formatBNToShortString(userShareData?.usdBalance || Zero, 18),
     ),
     tokens: poolData.tokens.map((coin) => {
       const token = TOKENS_MAP[coin.symbol]
@@ -104,7 +108,7 @@ function PoolOverview({
             <span>{`$${formattedData.reserve}`}</span>
           </div>
           {formattedData.volume && (
-            <div className="margin">
+            <div>
               <span className="label">{`${t("24HrVolume")}`}</span>
               <span>{formattedData.volume}</span>
             </div>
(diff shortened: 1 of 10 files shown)


src/components/PoolOverview.tsx
```diff
@@ -35,6 +35,15 @@ function PoolOverview({
           }
         : acc
     }, {} as Partial<Record<Partners, string>>),
+    apy: poolData.apy
```
hammeiam, line 38: generally prefer es6 string templates over concatenating strings. Also `.toFixed` returns a string so `String()` is unnecessary 
```
`${parseFloat(poolData.apy * 100).toFixed(4)}%`
```
0.123456 -> "12.3456%"

src/components/PoolOverview.tsx
```diff
@@ -91,6 +99,24 @@ function PoolOverview({
             <span className="label">{t("currencyReserves")}</span>
             <span>{`$${formattedData.reserve}`}</span>
           </div>
+          {formattedData.utilization && (
+            <div className="volume">
+              <span className="label">Utilization</span>
```
hammeiam, line 104: pls add these titles to our en.json and zh.json translation files. (zh entry can be null)

src/hooks/usePoolData.ts
```diff
@@ -33,15 +33,15 @@ export type Partners = "keep" | "sharedStake" | "alchemix"
 export interface PoolDataType {
   adminFee: BigNumber
   aParameter: BigNumber
-  apy: string // TODO: calculate
+  apy: string
```
hammeiam, line 36: can we make the new things you're adding `number` or `BigNumber` instead of strings? I realize the existing type said string but that was unused and incorrect.

src/components/PoolOverview.tsx
```diff
@@ -91,6 +99,24 @@ function PoolOverview({
             <span className="label">{t("currencyReserves")}</span>
             <span>{`$${formattedData.reserve}`}</span>
           </div>
+          {formattedData.utilization && (
+            <div className="volume">
```
hammeiam, line 103: add the `margin` class to each of these to add margin to the right and make the spacing correct

src/components/PoolOverview.tsx
```diff
@@ -35,6 +35,15 @@ function PoolOverview({
           }
         : acc
     }, {} as Partial<Record<Partners, string>>),
+    apy: poolData.apy
+      ? String((100 * parseFloat(poolData.apy)).toFixed(1)) + "%"
+      : "",
+    volume: poolData.volume
+      ? "$" + commify(parseFloat(poolData.volume).toFixed(0))
+      : "",
+    utilization: poolData.utilization
```
hammeiam, line 44: per the design, we can remove utilization from the `PoolOverview` and just include it in the `PoolInfoCard` component

src/components/PoolOverview.tsx
```diff
@@ -35,6 +35,15 @@ function PoolOverview({
           }
         : acc
     }, {} as Partial<Record<Partners, string>>),
```
hammeiam, line 37: This isn't feedback on your code, but could you change the `reserve` formatting to `commify(formatBNToString(poolData.reserve, 18, 0))` to get rid of the cents? Makes it cleaner :)

src/components/PoolOverview.tsx
```diff
@@ -91,6 +99,24 @@ function PoolOverview({
             <span className="label">{t("currencyReserves")}</span>
             <span>{`$${formattedData.reserve}`}</span>
           </div>
+          {formattedData.utilization && (
+            <div className="volume">
+              <span className="label">Utilization</span>
+              <span>{formattedData.utilization}</span>
+            </div>
+          )}
+          {formattedData.apy && (
```
hammeiam, line 108: Could you change the order of these so they're in the order APY, APR, Reserves, 24h Volume?

src/state/application.ts
```diff
@@ -5,6 +5,13 @@ interface GasPrices {
   gasFast?: number
   gasInstant?: number
 }
+interface SwapStats {
+  [swapAddress: string]: {
+    oneDayVolume: string
+    APY: string
```
hammeiam, line 11: nit: prefer making these lowercase to stay consistent with camelCase

src/pages/App.tsx
```diff
@@ -113,7 +114,11 @@ function GasAndTokenPrices({
   const fetchAndUpdateTokensPrice = useCallback(() => {
     fetchTokenPricesUSD(dispatch, chainId, library)
   }, [dispatch, chainId, library])
+  const fetchAndUpdateSwapStats = useCallback(() => {
+    void fetchSwapStats(dispatch)
+  }, [dispatch])
   usePoller(fetchAndUpdateGasPrice, BLOCK_TIME)
   usePoller(fetchAndUpdateTokensPrice, BLOCK_TIME * 3)
+  usePoller(fetchAndUpdateSwapStats, BLOCK_TIME * 3)
```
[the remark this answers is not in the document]
bobtista: blocktime is like 15sec, so 4blocks/min * 60min/hr  = 240blocks/hr, so we want to use 240*block_time, correct?

alphastorm (approved):

PR #507, opened 2021-06-25
PendingSwaps 1: timeline component

hammeiam:
This PR creates the timeline component for pendingSwaps, as well as corrects how pendingSwap data is represented (it previously incorrectly assumed that each pendingSwap would only have one Settlement)

(just the timeline, the rest has been split into other PRs)
Screen Shot 2021-06-24 at 9 31 14 AM


sandrasong (approved):

penandlim (approved):
LGTM

PR #514, opened 2021-06-28
Format pool figures better

hammeiam:
Screen Shot 2021-06-28 at 3 40 01 PM

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -37,6 +37,12 @@ export enum ChainId {
   // KOVAN = 42,
   HARDHAT = 31337,
 }
+export enum PoolTypes {
+  BTC,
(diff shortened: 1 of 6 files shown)


src/constants/index.ts
```diff
@@ -323,6 +329,7 @@ export type Pool = {
   poolTokens: Token[]
   isSynthetic: boolean
   addresses: { [chainId in ChainId]: string }
+  type: PoolTypes
```
sandrasong, line 332: Why does Token need poolType?

sandrasong (approved):
Looks good

penandlim:
image
Trailing 0 past the decimal point does not show up. Is this intended?

PR #516, opened 2021-06-29
Update transaction deadline settings from 10,30 min to 20,40 min

### LINKED ISSUE #513: Increase transaction deadline options to 20min and 40min

penandlim:
Changes the available options to 20 min and 40 min.
Sets the default placeholder for custom deadline to 20 min

Closes #513

### DIFF
--- a/src/state/user.ts
+++ b/src/state/user.ts
@@ -21,7 +21,9 @@ export enum Slippages {
 }
 
 export enum Deadlines {
+  Ten = "TEN",
   Twenty = "TWENTY",
+  Thirty = "THIRTY",
   Forty = "FORTY",
   Custom = "CUSTOM",
 }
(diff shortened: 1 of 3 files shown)


src/state/user.ts
```diff
@@ -21,8 +21,8 @@ export enum Slippages {
 }
 
 export enum Deadlines {
-  Ten = "TEN",
-  Thirty = "THIRTY",
+  Twenty = "TWENTY",
```
hammeiam, line 24: Caution that these keys are persisted as cookies. Trying to read deleted values when hydrating the reducer could cause issues

penandlim: Updated to keep the old values as well.

alphastorm (approved):

PR #558, opened 2021-07-29
sUSD Metapool

hammeiam:
- Incorporate the concept of a metapool across the app
- Add the susd metapool
- Refactor useContract to be slimmer
- Refactor routes to be derived from our pools list

All flows tested and working locally


src/constants/index.ts
```diff
@@ -238,9 +287,26 @@ export const D4_SWAP_TOKEN = new Token(
...
+const SUSD_CONTRACT_ADDRESSES: { [chainId in ChainId]: string } = {
+  [ChainId.MAINNET]: "0x57Ab1ec28D129707052df4dF418D58a2D46d5f51",
+  [ChainId.ROPSTEN]: "",
+  [ChainId.HARDHAT]: "0x0E801D84Fa97b50751Dbf25036d067dCf18858bF",
+}
+export const SUSD = new Token(
+  SUSD_CONTRACT_ADDRESSES,
+  18,
+  "sUSD",
+  "nusd",
```
alphastorm, line 304: s/nusd/susd

https://www.coingecko.com/en/coins/susd

src/hooks/usePoolData.ts
```diff
@@ -114,12 +120,25 @@ export default function usePoolData(
       )
         return
       const POOL = POOLS_MAP[poolName]
+      const effectivePoolTokens = POOL.underlyingPoolTokens || POOL.poolTokens
+      const isMetaSwap = POOL.metaSwapAddresses != null
+      let metaSwapContact = null as MetaSwap | null
```
alphastorm, line 125: s/Contact/Contract

alphastorm (commented):
coverage CI is broken at the moment

src/constants/index.ts
penandlim: The id required for API call is still nusd :S

penandlim:
Looks good but is there a way we can let users deposit with the USDv2 LP token instead? 
Current setup would make it so the exsiting LPs need to withdraw from USDv2 then redeposit to the susd meta pool.

penandlim:
image
Found an icon size issue - sUSD seems too big @sandrasong

sandrasong:
@penandlim Fixed the different icon size bug. The reason was the icons were different size. We used 18x18 in the previous design. But icons in current design are 24 x 24. I will create a ticket to update all icons.

alphastorm (approved):
lgtm, tested locally and everything works 🥳 

we should also add the option to withdraw wrapped as well

penandlim (approved):

PR #567, opened 2021-08-02
add withdrawal test flow

wlrd:
Setup cypress test for basic withdrawal. Includes combo and single selected withdrawals.


cypress/integration/withdrawal.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      cy.wait(3000)
+
+      // test single item
+      const tokens = poolTokens[poolName]
+      cy.get(".radio_wrapper .label").contains(tokens[0]).click()
+      cy.get(".tokenInput input").first().type("1")
+      cy.wait(500)
+      cy.get("button").contains("Withdraw").click()
+      cy.get("button").contains("Confirm Withdraw").click()
```
hammeiam, line 55: For each of these tests, could you add one addition assertion that the user's staked balance has decreased? Take a look at how this is done in the deposit test.

wlrd: just updated with that.

PR #602, opened 2021-08-21
BTC V2 pool launch

sandrasong:
Prepare the front end for BTC v2 pool.

Screen Shot 2021-08-20 at 5 12 46 PM


src/pages/Pools.tsx
```diff
@@ -121,44 +130,46 @@ function Pools(): ReactElement | null {
...
-                : 1
-            }
-          })
           .map(([poolProps, migrationPool]) => (
             <PoolOverview
               key={poolProps.name}
               {...poolProps}
               onClickMigrate={
-                migrationPool
+                migrationPool == STABLECOIN_POOL_V2_NAME
```
hammeiam, line 172: this shouldn't need to be changed since btc pool won't have a `migrationPool`

hammeiam (approved):

src/pages/Pools.tsx
sandrasong: good call, I forgot to change it back.

PR #613, opened 2021-09-03
Add wCUSD pool

hammeiam:


src/constants/index.ts
```diff
@@ -354,6 +390,20 @@ export const D4_SWAP_TOKEN = new Token(
 )
 
 // Stablecoins
+const TBTC_V2_CONTRACT_ADDRESSES: { [chainId in ChainId]: string } = {
```
alphastorm, line 393: nit: comment above is incorrect, this isn't a stablecoin

penandlim (approved):

PR #646, opened 2021-09-22
E2E Risk View

ug02fast:
Coverage for the Risk view.

image


.env
```diff
@@ -1,4 +1,4 @@
 REACT_APP_CHAIN_ID="1"
 REACT_APP_GOOGLE_ANALYTICS_MEASUREMENT_ID=""
 REACT_APP_NETWORK_URL="http://127.0.0.1:8545/"
-REACT_APP_NOTIFY_DAPP_ID="4ea9b565-dff8-4415-a65d-21c085a6110c"
\ No newline at end of file
+REACT_APP_NOTIFY_DAPP_ID="abc43422-2242-4292-8223-0a02687d627f"
```
ug02fast, line 4: @hammeiam what are your thoughts on this change?

package.json
```diff
@@ -94,6 +94,7 @@
     "build-storybook": "build-storybook -s public --no-dll",
     "cy:run": "cypress run",
     "cy:open": "cypress open",
+    "cy:startopen": "start-test 3000 'cypress open'",
```
ug02fast, line 97: need to start the server first when working in wsl2 environment

src/pages/Risk.tsx
```diff
@@ -12,15 +12,15 @@ function Risk(): ReactElement {
     <div className="riskpage">
       <TopMenu activeTab={t("risk")} />
       <div className="content">
-        <p>
+        <p data-cy="risk-intro">
```
ug02fast, line 15: using cypress recommended convention for targeting

.env
hammeiam: I don't think we should need to change this - were you runninng into errors with the old one?

src/pages/Risk.tsx
hammeiam: we also use `data-testid` in a few places. Since this is just a data attribute it shouldn't matter which we choose, but I think I'd prefer to just pick one or the other to be consistent.

.env
ug02fast: no, I just made the change right off the bat.

src/pages/Risk.tsx
ug02fast: i can change it to `data-testid` as it's only important that it's unique and use that going forward.

.env
ug02fast: I just confirmed that this older key `abc...` this error comes up: 
image

it doesn't happen when the key is updated.

hammeiam (approved):
Nice work!

PR #650, opened 2021-09-23
Implement hook to add token to metamask

adijo:
Implement hook to add Token to MetaMask. This hook is generic for any `Token` obj.

I've left out the `image` parameter for now

As there's no UI element currently using this functionality, I tested manually by adding a dummy button & couple of tokens (`STABLECOIN_SWAP_TOKEN` and `TBTC_SWAP_TOKEN`)

The `.eslintrc` is for the `@ts-ignore` that we need as ethers returns an incorrect type

Screen Shot 2021-09-23 at 4 37 06 PM

Screen Shot 2021-09-23 at 4 37 16 PM

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,6 @@
+
 {
-  "extends":  [
+  "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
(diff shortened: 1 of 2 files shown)


hammeiam (approved):
looks good! Can't wait to use it

PR #662, opened 2021-10-03
Notify users on contract errors

adijo:


adijo:
Sandra's commit got included somehow, will create another PR

Issue #691, opened 2021-11-09
[dApp / bug] Withdraw page side modal missing section dividers + background image?


zimh:
This is the deposit page (header nav > Pools > click any `deposit` button):
![image](https://user-images.githubusercontent.com/19593719/140940204-7884d443-41c6-44b9-b3c8-721750701fc9.png)

This is the withdraw page: (header nav > Pools > click any `withdraw` button):
![Screen Shot 2021-11-09 at 9 16 20 AM](https://user-images.githubusercontent.com/19593719/140940323-3f4e9eb1-3e2f-4760-8b08-d2b4c73a6015.png)

Notice:
* Right panel/modal, deposit page has red line separating sections, whereas withdraw page does not have this.
* Deposit page has seahorse in background, withdraw page does not have this

Rec fix:
* withdraw page should have red line separating all side panel/modal sections (between My Shares (not shown in screenshots) and NAME Pool, and then Currency reserves) -- basically make withdraw page side panel layout same as deposit page layout
* withdraw page should have same seahorse in bg as deposit page

ug02fast:
https://github.com/saddle-finance/saddle-frontend/pull/692

PR #692, opened 2021-11-09
Fix bug on withdraw page

nkrulikovsky:
## What does this PR do?

- Adding divider on right panel on withdraw page.
- Add background on withdraw page.

## Context

https://github.com/saddle-finance/saddle-frontend/issues/691

## Screenshot

image


sandrasong (approved):

src/components/WithdrawPage.tsx
```diff
@@ -82,7 +82,7 @@ const WithdrawPage = (props: Props): ReactElement => {
   const noShare = !myShareData || myShareData.lpTokenBalance.eq(Zero)
 
   return (
-    <div className={"withdraw " + classNames({ noShare: noShare })}>
+    <div className={classNames({ noShare: noShare, withdraw: true })}>
```
hammeiam, line 85: since the `noShare` class has been commented out of the scss, we can remove it here. The result is just `<div className="withdraw">`

ug02fast (approved):

PR #697, opened 2021-11-16
Enable L2, launch arbUSD

hammeiam:
- create arbUSD pool
- enable L2 flag
- update routes to redirect back to `/pools` if the user is in a pool then switches chains to one where that pool isn't supported 

I don't have any arb tokens so I'd like help testing


penandlim (approved):
LGTM

PR #713, opened 2021-12-01
Account detail, Advanced option test

nkrulikovskiy:
## What does this PR do?

- Account detail test.
- Advanced option test


cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,56 @@
...
+            .find("input")
+            .type(advancedOptionValues[index])
+            .should("have.value", advancedOptionValues[index])
+        })
+      } else {
+        cy.log("There is no table container")
+      }
+    })
+  }
+  it("input slipage value", () => {
```
ug02fast, line 53:
```suggestion
  it("input slippage value", () => {
```

src/components/AdvancedOptions.tsx
```diff
@@ -39,18 +39,19 @@ export default function AdvancedOptions(): ReactElement {
     (state: AppState) => state.application,
   )
 
+  console.log("advanced === >", advanced)
```
ug02fast, line 42:
```suggestion
```

src/components/AdvancedOptions.tsx
```diff
@@ -230,7 +230,7 @@ export default function AdvancedOptions(): ReactElement {
               <input
                 type="text"
                 className={classNames({
-                  [styles.selected]: gasPriceSelected === GasPrices.Custom,
```
ug02fast, line 233: What is the advantage of moving away from the `styles` class?

cypress/integration/accountDetail.test.ts
```diff
@@ -0,0 +1,16 @@
...
+    cy.visit(`${host}#/`)
+    cy.waitForReact()
+    cy.wait(3000)
+  })
+
+  it("renders account detail after click account avatar", () => {
+    cy.get(".walletStatus").should("exist")
+    cy.get(".walletStatus>button").click()
+    cy.get(".accountDetail").should("exist")
+    cy.get(".buttonGroup>button").contains("Change Account").click()
```
ug02fast, line 13: If these tests are passing then that's good. In the future try to move to directly targeting `data-testid` attributes. This will help in improving the robustness of E2E tests. Classes can become obfuscated, clash, and more and so can doing things like targeting elements.

As explained here: https://docs.cypress.io/guides/references/best-practices#How-It-Works

ug02fast (changes requested):

src/components/AdvancedOptions.tsx
```diff
@@ -1,3 +1,4 @@
+import "./AdvancedOptions.scss"
```
hammeiam, line 1: No, moving away from css modules is the wrong direction. CSS modules namespace classnames to prevent conflicts from other files.

I recommend following Arthur's suggesting of using `data-testid` to help with element targeting

src/components/AdvancedOptions.tsx
hammeiam: makes testing a little easier, but I don't think it's worth it

hammeiam (changes requested):
We should use `data-testid` instead of regressing from css modules to class names

src/components/AdvancedOptions.tsx
ug02fast: Yup just add `data-testid` to components needing targeting and pass reasonable name through as the value to that attribute.

nkrulikovskiy:
Oh, Got it.
Next time, I'll use `data-testid` . I think this method will work well for `styled-component` as well.

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    cy.contains(poolName)
+      .parents(".poolOverview")
+      .within(() => {
+        cy.get("button").contains("Withdraw").click()
+      })
+    cy.wait(10000)
+
+    cy.get("[data-testid=optionContainer]").should("exist")
+    cy.get("[data-testid=optionContainer]").then((advancedOptions) => {
+      if (advancedOptions.find("[dat-testid=tableContainer]").length > 0) {
```
ug02fast, line 28: `data-testid`

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        cy.get("[dat-testid=tableContainer]").then(($table) => {
```
ug02fast, line 29: same here

src/components/AdvancedOptions.tsx
```diff
@@ -68,12 +69,16 @@ export default function AdvancedOptions(): ReactElement {
       </span>
       <div className={styles.divider}></div>
       <div
+        dat-testid="tableContainer"
```
ug02fast, line 72: `data-testid`

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    cy.visit(`${host}#/pools`)
+    cy.wait(3000)
+    cy.contains(poolName)
+      .parents(".poolOverview")
+      .within(() => {
+        cy.get("button").contains("Withdraw").click()
+      })
+    cy.wait(10000)
+
+    cy.get("[data-testid=optionContainer]").should("exist")
```
ug02fast, line 26: This is a good start. These names will work now but if there are other components with `option container` in the future then this test could require maintenance to in case it targets the other "option container", it could clash with this targeting. Something like `advOptionContainer` could be a better name just to scope it just a little more to this component.

src/components/AdvancedOptions.tsx
```diff
@@ -86,7 +91,7 @@ export default function AdvancedOptions(): ReactElement {
           </div>
         </div>
         <div className={styles.parameter}>
-          <div className={styles.inputGroup}>
+          <div data-testid="inputGroup" className={styles.inputGroup}>
```
ug02fast, line 94: Same thing as above. The naming for this would be better as something like `maxSlippageInputGroup`.

src/components/AdvancedOptions.tsx
```diff
@@ -130,7 +135,7 @@ export default function AdvancedOptions(): ReactElement {
...
```
ug02fast, line 138: Same thing as above. The naming for this would be better as something like `deadlineInputGroup`.

src/components/AdvancedOptions.tsx
```diff
@@ -191,7 +196,7 @@ export default function AdvancedOptions(): ReactElement {
           </div>
         </div>
         <div className={styles.parameter} style={{ display: "none" }}>
-          <div className={styles.inputGroup}>
+          <div data-testid="inputGroup" className={styles.inputGroup}>
```
ug02fast, line 199: etc.

src/components/AdvancedOptions.tsx
```diff
@@ -68,12 +69,16 @@ export default function AdvancedOptions(): ReactElement {
...
       <div
+        dat-testid="tableContainer"
         className={classNames(styles.tableContainer, {
           [styles.show]: advanced,
         })}
       >
         <div className={styles.parameter}>
-          <div className={styles.infiniteApproval}>
+          <div
+            data-testid="infiniteApproval"
```
ug02fast, line 79: Good name here, explicit. Maybe even naming it `infiniteApprovalContainer` would make it more specific.

src/components/AdvancedOptions.tsx
ug02fast, line 72: misspelled

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+                    const buttonClass = $afterClick.attr("class")
+                    expect(buttonClass).to.match(/selected/)
+                  })
+              })
+            cy.wrap($inputGrop)
+              .find("input")
+              .type(advancedOptionValues[index])
+              .should("have.value", advancedOptionValues[index])
+          } else {
+            cy.log("input group is invisibel")
```
ug02fast, line 60: invisible*

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+          if (!isTableVisible) {
+            cy.get("[data-testid=title]").click()
+          }
+        })
+        cy.get("[data-testid=infiniteApproval]")
+          .find('[type="checkbox"]')
+          .check({ force: true })
+
+        cy.get("[data-testid=inputGroup]").each(($inputGrop, index) => {
```
ug02fast, line 41: inputGroup*

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+          const display = $inputGrop.parent().css("display")
```
ug02fast, line 42: inputGroup*

cypress/integration/advancedOption.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+          if (display !== "none") {
```
ug02fast, line 44: We should be caution of using conditional statements in Cypress tests.

https://docs.cypress.io/guides/core-concepts/conditional-testing#The-DOM-is-unstable

Maybe adding data tag to "inputGroup" parent (if possible) then checking like

`cy.get(['data-testid=inputGroupContainer']).should('not.exist')` Something like this could work

ug02fast (changes requested):
Good progress so far. There are a few things we have to do in order to keep e2e tests maintainable and robust long term.

ug02fast (approved):

PR #720, opened 2021-12-09
Nikita/modal provider

nkrulikovsky:
- modal provider.
- Replace the review modal on the withdraw with new modal provider.


src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+interface ModalsContext {
+  content?: React.ReactNode
+  isOpen?: boolean
+  onPresent: (content: React.ReactNode, key?: string) => void
+  onClose: () => void
+}
+
+export const Context = createContext<ModalsContext>({
+  onPresent: () => undefined,
```
ug02fast, line 17: `() => void` describes this action more precisely since technically you aren't returning undefined but rather nothing is happening, hence void. Also leverages TS here.

src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onClose: () => undefined,
```
ug02fast, line 18: `() => void` describes this action more precisely since technically you aren't returning undefined but rather nothing is happening, hence void. Also leverages TS here.

src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+    (modalContent: React.ReactNode, key?: string) => {
+      setModalKey(key)
+      setContent(modalContent)
+      setIsOpen(true)
+    },
+    [setContent, setIsOpen, setModalKey],
+  )
+
+  const handleClose = useCallback(() => {
+    setContent(undefined)
```
ug02fast, line 36: Setting to undefined can be risky as it can lead to a TypeError and crash the app. Is it possible to set content to something like `<div />` or `null` or something that empties that content without using `undefined`?

src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+import styles from "./ModalProvider.module.scss"
+
+interface ModalsContext {
+  content?: React.ReactNode
+  isOpen?: boolean
+  onPresent: (content: React.ReactNode, key?: string) => void
+  onClose: () => void
+}
+
+export const Context = createContext<ModalsContext>({
```
ug02fast, line 16: `export const ModalContext = createContext<ModalContextProps>({` What do you think about this naming and convention?

src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
+import React, {
+  PropsWithChildren,
+  createContext,
+  useCallback,
+  useState,
+} from "react"
+import styles from "./ModalProvider.module.scss"
+
+interface ModalsContext {
```
ug02fast, line 9: `ModalContextProps` thoughts on this naming convention?

src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+      }}
+    >
+      {children}
+      {isOpen && (
+        <div className={styles.modalWrapper}>
+          <div className={styles.modalBackdrop} onClick={handleClose} />
+          {React.isValidElement(content) &&
+            React.cloneElement(content, {
+              onClose: handleClose,
+            })}
```
ug02fast, line 56: Can you use something like this here instead? It's hard to understand what 'content' is here, can it be more explicit of a name or is this value too generic for that? What do you think about calling it ModalBody or ModalContent? Does it work without using React.cloneElement?

```
<ModalContent onClose={handleClose} />
```

src/providers/ModalsProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+})
+
+const ModalsProvider: React.FC = ({ children }: PropsWithChildren<unknown>) => {
+  const [isOpen, setIsOpen] = useState(false)
+  const [content, setContent] = useState<React.ReactNode>()
+  const [, setModalKey] = useState<string>()
+
+  const handlePresent = useCallback(
+    (modalContent: React.ReactNode, key?: string) => {
+      setModalKey(key)
```
ug02fast, line 28: What is `key` being used for here?

src/hooks/useModal.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Context } from "../providers/ModalsProvider"
+
+const useModal = (
+  modal: React.ReactNode,
+  key?: string,
+): [() => void, () => void] => {
+  const { onClose, onPresent } = useContext(Context)
+
+  const handlePresent = useCallback(() => {
+    onPresent(modal, key)
```
ug02fast, line 11: What does Present mean? Does that mean like "Show"?

ug02fast (changes requested):

src/hooks/useModal.ts
nkrulikovskiy: Yes, it means "show modal".

src/providers/ModalsProvider.tsx
nkrulikovskiy: Right, I have tried to give void, but it shows me type error. That is the reason why I gave undefined instead of void 
Screen Shot 2021-12-10 at 2 02 58 PM

src/providers/ModalsProvider.tsx
nkrulikovskiy: I put this value to identify modal in the case of display modal on the another modal. But our case it doesn't require, so I removed.

src/components/Modal/ModalAction.tsx
```diff
@@ -0,0 +1,18 @@
...
+import style from "./ModalAction.module.scss"
+
+const ModalActions: React.FC = ({ children }: PropsWithChildren<unknown>) => {
+  const childrenLength = React.Children.toArray(children).length
+  return (
+    <div className={style.modalActions}>
+      {React.Children.map(children, (child, i) => (
+        <>
+          <div className={style.modalAction}>{child}</div>
+          {i < childrenLength - 1 && <div />}
```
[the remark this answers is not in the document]
nkrulikovskiy: On material UI, there is ModalAction component which is have the buttons such as `confirm`, `cancel`, etc.
But we are having already modal content which have these buttons, so I think ModalAction is not required. I removed.

src/providers/ModalsProvider.tsx
ug02fast: I mis-read. This should be `() => {}` which more closely represent a `noop`

src/hooks/useModal.ts
ug02fast: Does using "show" cause clashing of terms elsewhere? Present technically works by definition but a little strange.

src/components/WithdrawPage.tsx
```diff
@@ -74,13 +76,23 @@ const WithdrawPage = (props: Props): ReactElement => {
...
-  const onSubmit = (): void => {
-    setCurrentModal("review")
+  const [openConfirmModal, closeConfirmModal] = useModal(
+    <Modal>
+      <ConfirmTransaction />
+    </Modal>,
+  )
+  const onConfirm = async (): Promise<void> => {
+    openConfirmModal()
+    logEvent("withdraw", (poolData && { pool: poolData?.name }) || {})
```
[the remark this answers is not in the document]
ug02fast: Oh this is just a logEvent, nbd.

src/providers/ModalsProvider.tsx
ug02fast: Update: `() => {}`

ug02fast (changes requested):
Added a couple other things. Will this provider work with material UI API?

src/providers/ModalsProvider.tsx
nkrulikovskiy: oh

nkrulikovskiy: Just remembered. when I put the `()=>{}` for initializing the function, it gives me the eslint error.
Screen Shot 2021-12-10 at 9 36 12 PM

nkrulikovskiy:
This ModalProvider will work with MUI component API.
Currently MUI don't have ModalProvider, and they only have Modal, ModalTitle, ModalAction. So, we can use their component on this provider without any issue if we use MUI library.

ug02fast (approved):

PR #721, opened 2021-12-10
add local test environment

nkrulikovskiy:
- After adding tally wallet, cypress don't work because there is no metamask on test envrionment.

### DIFF
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -21,57 +21,35 @@ export function useEagerConnect(): boolean {
 
   useEffect(() => {
     if (process.env.NODE_ENV === "production") {
-      if (window.ethereum) {
-        if (window.ethereum.isTally) {
-          void injectedTallyProvider.isAuthorized().then((isAuthorized) => {
-            if (isAuthorized) {
-              activate(injectedTallyProvider, undefined, true).catch(() => {
-                setTried(true)
-              })
-            } else {
-              if (isMobile && window.ethereum) {
-                activate(injectedTallyProvider, undefined, true).catch(() => {
-                  setTried(true)
-                })
-              } else {
-                setTried(true)
-              }
-            }
-          })
-        } else if (window.ethereum?.isMetaMask) {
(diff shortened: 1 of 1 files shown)


src/hooks/index.ts
```diff
@@ -20,40 +20,60 @@ export function useEagerConnect(): boolean {
...
-            activate(injectedTallyProvider, undefined, true).catch(() => {
-              setTried(true)
-            })
-          } else {
-            if (isMobile && window.ethereum) {
+    if (process.env.NODE_ENV === "production") {
+      if (window.ethereum) {
+        if (window.ethereum.isTally) {
+          void injectedTallyProvider.isAuthorized().then((isAuthorized) => {
+            if (isAuthorized) {
```
ug02fast, line 27: can imply `window.ethereum` is true from above

```
if (isAuthorized || isMobile) { ...
```

src/hooks/index.ts
```diff
@@ -20,40 +20,60 @@ export function useEagerConnect(): boolean {
...
-          if (isAuthorized) {
-            activate(injectedMetaMaskProvider, undefined, true).catch(() => {
-              setTried(true)
-            })
-          } else {
-            if (isMobile) {
+          })
+        } else if (window.ethereum?.isMetaMask) {
+          void injectedMetaMaskProvider.isAuthorized().then((isAuthorized) => {
+            if (isAuthorized) {
```
ug02fast, line 43: same as above `if (isAuthorized || isMobile)`

src/hooks/index.ts
```diff
@@ -20,40 +20,60 @@ export function useEagerConnect(): boolean {
...
+              }
             }
-          }
-        })
+          })
+        }
       }
+    } else {
+      void injectedMetaMaskProvider.isAuthorized().then((isAuthorized) => {
+        if (isAuthorized) {
```
ug02fast, line 63: same as above

ug02fast (changes requested):

PR #722, opened 2021-12-10
Investor-Employee SDL Claim UI

ug02fast:
Valid Beneficiary with claimable:
image

When using invalid address to claim:
image

UI leveraging existing components strictly for Investors/Employees to claim SDL.

Will be located at `saddle.exchange.com/#/vesting-claim`


src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+      notifyCustomError({
+        ...(err as Error),
+        message: "Unable to claim vested tokens",
+      })
+    }
+  }
+
+  return (
+    <>
+      <TopMenu activeTab={t("risk")} />
```
nkrulikovskiy: On this page, does we need to active risk?

nkrulikovskiy (changes requested):

src/pages/VestingClaim.tsx
ug02fast: I just changed this to n/a to swallow that logic.

sandrasong:
Look good. One small copywrite improvement could be this. 

team claim

ug02fast:
> Looks good Art!
> 
> One small copywrite improvement could be this. team claim
> 
> The sentence `**** of ****** remaining tokens` + much smaller size `total claimable SDL` is a bit confusing to me.

image
How is this?

sandrasong:
Thanks for changing, the ux copywriting is clear! 

Could you change the `+Add token button` alignment? Now that button feels like related to Claimable token. 

investor claim2

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+function VestingClaim(): ReactElement {
+  const { t } = useTranslation()
+  const { account, chainId, library } = useActiveWeb3React()
+  const { addToken, canAdd } = useAddTokenToMetamask({
+    ...SDL_TOKEN,
+    icon: `${window.location.origin}/logo.svg`,
+  })
+
+  const [isValidBene, setIsValidBene] = useState(false)
```
hammeiam, line 27: nit: prefer full words

src/pages/App.tsx
```diff
@@ -75,6 +76,7 @@ export default function App(): ReactElement {
                   ))}
                   <Redirect from="/pools/:route/:action" to="/pools" />
                   <Route exact path="/risk" component={Risk} />
+                  <Route exact path="/vesting-claim" component={VestingClaim} />
```
hammeiam, line 79: optimization: Could you make Vesting (and Risk, while you're at it) a lazy-loaded route? Should be a 1 line change 
https://reactjs.org/docs/code-splitting.html#route-based-code-splitting

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const [claimableVestedAmount, setClaimableVestedAmount] = useState("0.0")
+  const [remainingAmount, setRemainingAmount] = useState("0.0")
+  const [vestingContract, setVestingContract] = useState<Vesting>()
+
+  useEffect(() => {
+    const fetchBeneficiaries = async () => {
+      if (!library || !chainId) return
+      const provider = new ethers.providers.Web3Provider(
```
hammeiam, line 35: Prefer to use `library` from `useActiveWeb3React` above instead of creating a new provider

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+        window.ethereum as never,
+      )
+      const SDLContract = new EthersContract(
+        SDL_TOKEN_ADDRESSES[chainId],
+        SDL_TOKEN_ABI,
+        provider.getSigner(),
```
hammeiam, line 41: Prefer to use `account` from `useActiveWeb3React` - check out the `useContract` hook or https://github.com/saddle-finance/saddle-frontend/blob/master/src/hooks/usePoolData.ts#L166 for examples.

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+      ) as Sdl
+      const vestingContractDeployedFilter = SDLContract.filters.VestingContractDeployed(
+        null,
+        null,
+      )
+      try {
+        const events = await SDLContract.queryFilter(
+          vestingContractDeployedFilter,
+        )
+        const currentBeneficiary = events.filter(
```
hammeiam, line 51: nit: `currentAccountEvents` or `currentBeneficiaryEvents`

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+          (event) => event.args?.beneficiary === account,
```
hammeiam, line 52: It doesn't seem like it, but could casing be an issue?

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+        )
+        setIsValidBene(currentBeneficiary.length > 0)
+        if (isValidBene) {
+          const vestingContract = new EthersContract(
+            currentBeneficiary[0].args?.vestingContract,
+            INVESTOR_EMPLOYEE_VESTING_CONTRACT_ABI,
+            provider.getSigner(),
```
hammeiam, line 59: same comment as in above contract

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+function VestingClaim(): ReactElement {
+  const { t } = useTranslation()
+  const { account, chainId, library } = useActiveWeb3React()
+  const { addToken, canAdd } = useAddTokenToMetamask({
+    ...SDL_TOKEN,
+    icon: `${window.location.origin}/logo.svg`,
+  })
+
+  const [isValidBene, setIsValidBene] = useState(false)
+  const [claimableVestedAmount, setClaimableVestedAmount] = useState("0.0")
```
hammeiam, line 28: Prefer to keep raw `BigNumber`'s in state, and format them in UI. The reason is that we may need to do math or display multiple formats with those numbers, so we want to keep them in their pure form. Therefore your default value for both of these is `Zero` from `import {Zero} from "@ethersproject/constants"`

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <div className={styles.logoWrapper}>
+          <div className={styles.logo}>
+            <img src={logo} />
+          </div>
+        </div>
+        <div className={styles.mainContent}>
+          {isValidBene ? (
+            <>
+              <div className={styles.tokenBalance}>
+                {claimableVestedAmount} of {remainingAmount} remaining tokens
```
hammeiam, line 132: TODO: translate. grep for `stepAOfB` for how to insert variables into translations

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+        notifyCustomError({
+          ...(err as Error),
+          message: "Unable to query for vesting contracts",
+        })
+      }
+    }
+    void fetchBeneficiaries()
+  }, [account, chainId, library, isValidBene])
+
+  const onClaimClick: () => void = async () => {
```
hammeiam, line 102: Confirm that this doesn't need to be wrapped in a `useCallback`

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    onClick={() => addToken()}
+                  />
+                )}
+              </div>
+              <div className={styles.tokenBalanceHelpText}>
+                {t("totalClaimableSDL")}
+              </div>
+              <div className={styles.tokenBalanceHelpText}>
+                Vesting Contract:{" "}
+                <a
```
hammeiam, line 146: nit: hide the link if we don't have a vesting contract address

hammeiam (commented):
Looking really good so far! I just have some small behind the scenes changes, but overall this is great.

src/pages/VestingClaim.tsx
ug02fast: The containing div won't render if the beneficiary isn't valid which won't render this div as a result.

src/pages/VestingClaim.tsx
ug02fast: Wrapping this function in a `useCallback` isn't recommended as explained in https://dmitripavlutin.com/dont-overuse-react-usecallback/

src/pages/VestingClaim.tsx
ug02fast: I was struggling with finding a decent name here. Yours are more concise.

src/pages/VestingClaim.tsx
ug02fast: both addresses here are non-lowercased (checksummed). good call out, good to dbl check.

src/pages/VestingClaim.tsx
ug02fast: Should all the displayed text be translated?

ug02fast:
> Thanks for changing, the ux copywriting is clear!
> 
> Could you change the `+Add token button` alignment? Now that button feels like related to Claimable token.
> 
> investor claim2

Updated:
image

src/pages/App.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
ug02fast, line 2: @hammeiam I wasn't able to figure out how to resolve these TS errors for the loadable library which is necessary for code splitting. Do you happen to know how to resolve it for lines 35 & 36?

hammeiam: I think we can just use React.lazy since we're not SSR

src/pages/VestingClaim.tsx
hammeiam: I'm not actually sure how many of our investors are ZH-first/only, translating is probably not a high-pri. It is necessary for public-facing features though.

src/pages/VestingClaim.tsx
```diff
@@ -0,0 +1,198 @@
...
+      )
+      try {
+        const events = await SDLContract.queryFilter(
+          vestingContractDeployedFilter,
+        )
+        const currentBeneficiaryEvents = events.filter(
+          (event) => event.args?.beneficiary === account,
+        )
+        setIsValidBeneficiary(currentBeneficiaryEvents.length > 0)
+        if (isValidBeneficiary) {
```
hammeiam, line 57: I think this will cause this effect to be run once more than expected, since `isValidBeneficiary` is both set by this effect and a dependency of it. To fix, remove it as an effect dep and just store / read `currentBeneficiaryEvents.length > 0` as a variable within the effect.

ug02fast: nice find, this doesn't need to be an effect dep

PR #723, opened 2021-12-10
Fix test breaking due to the isMetaMask on the cypress/support/commands.ts

nkrulikovskiy:
- Fix test breaking

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -53,7 +53,7 @@ class CustomizedBridge extends _Eip1193Bridge {
         error = new Error("Something went wrong")
       }
       if (isCallbackForm) {
-        callback(error, result ? { result } : null )
+        callback(error, result ? { result } : null)
       } else {
         return result ? Promise.resolve(result) : Promise.reject(error)
       }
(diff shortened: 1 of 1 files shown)


ug02fast (approved):

cypress/support/commands.ts
```diff
@@ -94,12 +96,16 @@ Cypress.Commands.overwrite("visit", (original, url, options) => {
...
+          name: "local",
+          chainId: Cypress.env("NETWORK_ID"),
+        })
+        const signer = new Wallet(
+          Cypress.env("PRIVATE_TEST_WALLET_PK"),
+          provider,
         )
-        const signer = new Wallet(Cypress.env("PRIVATE_TEST_WALLET_PK"), provider)
         win.ethereum = new CustomizedBridge(signer, provider)
+        win.ethereum.isMetaMask = true
```
hammeiam, line 108: Ideally we'd only change this property for the test that requires it, but this is fine for now.

cypress/support/commands.ts
```diff
@@ -62,7 +62,9 @@ class CustomizedBridge extends _Eip1193Bridge {
       return wrapResponse([Cypress.env("PRIVATE_TEST_WALLET_ADDRESS")])
     }
     if (method === "eth_chainId") {
-      return wrapResponse(`0x${Cypress.env("NETWORK_ID").toString(16).toUpperCase()}`)
+      return wrapResponse(
+        `0x${Cypress.env("NETWORK_ID").toString(16).toUpperCase()}`,
```
hammeiam, line 66: Please change your IDE line length to be 80 to avoid future unnecessary changes like this in the future

hammeiam (approved):

cypress/support/commands.ts
nkrulikovskiy: Oh, I am having already 80 for IDE line length.

PR #740, opened 2021-12-21
Added Ens name and avatar support

### LINKED ISSUE #739: Show ENS name and avatar in UI if owned
It would be great if we show the ENS name and avatar if the user has it

image

sasicodes:
closes #739 

cc @hammeiam


src/hooks/useENS.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type ReturnType = { ensName: string | null }
+
+export const useENS = (address: string | null | undefined): ReturnType => {
+  const [ensName, setENSName] = useState<string | null>(null)
+
+  useEffect(() => {
+    async function resolveENS() {
+      if (address && ethers.utils.isAddress(address)) {
+        const provider = new ethers.providers.JsonRpcProvider(
```
hammeiam, line 12: Prefer to use provider library from `const { library } = useActiveWeb3React()` instead of using Env variable

src/hooks/useENS.ts
```diff
@@ -0,0 +1,24 @@
...
+          process.env.REACT_APP_NETWORK_URL,
+        )
+        const ensName = await provider.lookupAddress(address)
+        if (ensName) setENSName(ensName)
+      }
+    }
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
hammeiam, line 19: You could add a `.catch(console.error)` so you don't need this comment

sasicodes:
@hammeiam Resolved comments 🤗

package.json
```diff
@@ -4,9 +4,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@davatar/react": "^1.8.1",
```
hammeiam, line 7: You'll also need to update the package-lock.json for this to build correctly

sasicodes: Done

sasicodes:
cc @hammeiam

hammeiam (approved):

hammeiam:
Thank you for your contribution!

PR #742, opened 2021-12-22
UI Fix for BTC Migrations

ug02fast:
BTC based pools will show the correct currency.

image

On hold, waiting for the new GeneralizedMigrator contract.

### DIFF
--- a/src/components/ReviewMigration.tsx
+++ b/src/components/ReviewMigration.tsx
@@ -6,6 +6,7 @@ import { commify, formatBNToString } from "../utils"
 import { AppState } from "../state/index"
 import { BigNumber } from "@ethersproject/bignumber"
 import Button from "./Button"
+import { PoolName } from "../constants"
 import Warning from "./Warning"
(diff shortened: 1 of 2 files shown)


src/components/ReviewMigration.tsx
```diff
@@ -56,7 +60,7 @@ function ReviewMigration({
               {commify(
                 formatBNToString(migrationAmount || BigNumber.from("0"), 18, 2),
               )}{" "}
-              saddleUSD
+              {isBTCMigration ? "saddleBTC" : "saddleUSD"}
```
hammeiam, line 63: Why can't we just display the pool name as it is? Or get the pools lpToken name? Those sound more correct, esp since we may have an eth pool that gets migrated

ug02fast: Sounds good to me

hammeiam: Nice find btw

hammeiam (approved):

Issue #741, opened 2021-12-21
App doesn't load when there is no metamask.


nkrulikovskiy:
- When there is no metamask or browser is on the incognito, app shows blank screen.

PR #743, opened 2021-12-22
Fix blank page when there is no metamask

nkrulikovskiy:
## What does this PR do?

- Now blank page is appeared  when there is no metamask.

## Context

#741 

## screenshot
Screen Shot 2021-12-22 at 11 58 54 AM

### DIFF
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -54,6 +54,8 @@ export function useEagerConnect(): boolean {
           }
         })
       }
+    } else {
+      setTried(true)
     }
   }, [activate]) // intentionally only running on mount (make sure it's only mounted once :))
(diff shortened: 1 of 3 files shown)


ug02fast:
Did you essentially wrap the pool calls in a try-catch?

hammeiam (approved):
We should still be able to read from the network given that we're using infura... this is a patch but it seems like there's a deeper issue here

nkrulikovskiy:
> Did you essentially wrap the pool calls in a try-catch? @nkrulikovskiy

Yes, And I gave the true to `tried` when there is no metamask on the useEagerConnect.

ug02fast (approved):

PR #745, opened 2021-12-22
frax-ArbUSDv2 Pool

ug02fast:
image

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -499,9 +499,9 @@ export const ARB_USD_SWAP_TOKEN = new Token(
 export const FRAX_ARB_USD_SWAP_V2_TOKEN = new Token(
   FRAX_ARB_USD_SWAP_V2_TOKEN_CONTRACT_ADDRESSES,
   18,
-  "saddleArbUSD-V2",
-  "saddlearbUSD-v2",
-  "Saddle nUSD/MIM/USDC/USDT",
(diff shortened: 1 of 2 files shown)


src/constants/index.ts
```diff
@@ -918,6 +941,16 @@ export const POOLS_MAP: PoolsMap = {
...
   },
+  [FRAX_ARB_USD_POOL_V2_NAME]: {
+    name: FRAX_ARB_USD_POOL_V2_NAME,
+    addresses: FRAX_ARB_USD_SWAP_V2_ADDRESSES,
+    lpToken: FRAX_ARB_USD_SWAP_V2_TOKEN,
+    poolTokens: FRAX_ARB_USD_POOL_V2_TOKENS,
+    isSynthetic: false,
+    type: PoolTypes.USD,
+    route: "frax-arbusdv2",
+    rewardPids: buildPids({ [ChainId.ARBITRUM]: 1 }),
```
hammeiam, line 952: double check with John on this, but I believe this pool is not yet in the arbitrum reward contract. I think it would be 2 (not 0-indexed) when added

hammeiam (approved):
Remove reward PID and you're good to go

src/constants/index.ts
ug02fast: John says it's going to be managed by Frax team on staking.frax.finance.

Issue #700, opened 2021-11-16
Show wrong network error message when trying to connect with unsupported networks


penandlim:

zimh:
was gonna create a new ticket but maybe the fix for this can also include fix for weird UI? Or LMK if better to create a separate ticket @sandrasong 

Issue (2): 
- When user is connected to an unsupported network, the UI shows "Ethereum" with green dot
![image](https://user-images.githubusercontent.com/19593719/145447594-8307eb21-c6df-4692-a7fc-a4dedb0c8bc9.png)
- When user is connected to an unsupported network and user attempts to switch to a network using the UI, this error shows:
![image](https://user-images.githubusercontent.com/19593719/145447682-5e0788da-64db-4653-a5bc-5e5a143e8c20.png)

Steps to repro:
- Set metamask to Avalanche (or another unsupported network)
- Observe that UI shows Ethereum with green dot
- Click on Ethereum > Network > Arbitrum
- Observe that nothing happens on the UI (and an error msg is sent in console)

Ideal solution: 
- when user is connected to an unsupported network, the UI should show a red dot instead of green
- when user is connected to an unsupported network and attempts to switch to a supported network, it should either successfully switch networks, or display a message to the user that they need to change network using their wallet first

sandrasong:
Design task addd to this ticket: 
- [ ] Create the wrong network notification

PR #753, opened 2021-12-28
Wrong network modal

nkrulikovskiy:
## What does this PR do?

- Display modal on wrong network.
- Change green dot  to red on Network display.

## Context 
#700 

## Screenshot
- Light:
Screen Shot 2021-12-29 at 12 51 05 AM

- Dark:
Screen Shot 2021-12-29 at 12 51 37 AM


src/components/WrongNetworkModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default function WrongNetworkModal({ open }: Props): ReactElement {
+  const { error } = useWeb3React<Web3Provider>()
+  const isUnsupportChainIdError = error instanceof UnsupportedChainIdError
+
+  return (
+    <Dialog open={open ?? isUnsupportChainIdError} maxWidth="xs">
+      <Box padding={3}>
+        <Typography variant="h3" textAlign="center" mb={3}>
```
hammeiam, line 17: Love to see this MUI!

src/components/WrongNetworkModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+          &#129325;
+        </Typography>
+        <p>
+          Opps, looks like you&apos;re trying to use Saddle while your wallet is
```
hammeiam, line 21: 1) please add translation to en.json
2) "Opps" should be "Oops"

src/theme.ts
```diff
@@ -102,8 +115,29 @@ export const darkTheme = createTheme({
...
+        },
+      },
+    },
+    MuiDialog: {
+      styleOverrides: {
+        root: {
+          borderRadius: 10,
+        },
+        paper: {
+          borderRadius: 10,
```
hammeiam, line 131: Prefer to have light and dark themes inherit all non-color overrides (like borders, spacings, fonts, etc) from a common theme, rather than duplicating code like this.

hammeiam (approved):

src/theme.ts
nkrulikovskiy: Good idea!

public/locales/zh.json
```diff
@@ -184,6 +184,7 @@
   "withdrawPercentage": "提取的百分比",
   "withdrawSynth": null,
   "withdrewNToken": null,
+  "wrongNetworkContent": null,
```
hammeiam, line 187: While we have added `null` translations to `zh.json` for other strings, I learned recently that this is not the correct approach. It is better *not* to make a null entry so that it defaults to English. 
I will make a PR to remove other null translations

nkrulikovskiy: Oh, got it.

Issue #746, opened 2021-12-23
Update Arbitrum reward logic


hammeiam:
As per this thread (https://saddle-finance.slack.com/archives/C02P1QQ1JBU/p1640182198014800):

- SDL rewards are not claimable on arbitrum
- The SDL retroactive airdrop was only for mainnet
- The claim UI on Arbitrum is just meant to be a preview of SDL rewards

Given this, we want to:
- Disable claiming rewards on Arbitrum, but still show users the rewards that are accumulating
- Add info clearly explaining the situation. Something like "SDL is currently only deployed on Ethereum Mainnet and is not yet claimable on this chain. We display the amount that will be claimable once SDL is available on this network. See this post for more information." cc zim for faq link
- Possibly hide the retro airdrop section on arbitrum, though it should be 0 anyway since the retro contract doesn't exist on that chain

zimh:
FAQ added https://docs.saddle.finance/saddle-faq#why-cant-i-claim-my-sdl-on-arbitrum

PR #754, opened 2021-12-29
hide retro airdrop section

nkrulikovsky:
## What does this PR do?

- Disable claiming rewards on Arbitrum.
- Add explaining the reason of claiming.

## Context

https://github.com/saddle-finance/saddle-frontend/issues/746
https://saddle-finance.slack.com/archives/C02P1QQ1JBU/p1640182198014800

## Screenshot
Screen Shot 2022-01-03 at 4 05 40 PM

### DIFF
--- a/src/components/TokenClaimModal.tsx
+++ b/src/components/TokenClaimModal.tsx
@@ -18,7 +18,6 @@ import {
 import AddCircleOutlineIcon from "@mui/icons-material/AddCircleOutline"
 import { BigNumber } from "@ethersproject/bignumber"
 import Button from "./Button"
-// import{ Link } from "react-router-dom"
 import { RewardsBalancesContext } from "../providers/RewardsBalancesProvider"
 import { Zero } from "@ethersproject/constants"
 import logo from "../assets/icons/logo.svg"
--- a/src/components/TokenClaimModal.tsx
+++ b/src/components/TokenClaimModal.tsx
@@ -120,7 +120,11 @@ export default function TokenClaimModal(): ReactElement {
               )}
             </>
           )}
-          {!isClaimableNetwork && <div>{t("desableRewardContent")}</div>}
+          {!isClaimableNetwork && (
+            <p style={{ whiteSpace: "pre-line" }}>
+              {t("desableRewardContent")}
+            </p>
+          )}
           {allPoolsWithRewards.map((pool, i, arr) => (
             <>
               <ClaimListItem
--- a/src/components/TokenClaimModal.tsx
+++ b/src/components/TokenClaimModal.tsx
@@ -14,12 +14,13 @@ import {
   useRetroactiveVestingContract,
 } from "../hooks/useContract"
 
+import AddCircleOutlineIcon from "@mui/icons-material/AddCircleOutline"
 import { BigNumber } from "@ethersproject/bignumber"
 import Button from "./Button"
+import { IconButton } from "@mui/material"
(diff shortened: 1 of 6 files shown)


src/components/TokenClaimModal.tsx
```diff
@@ -111,6 +118,23 @@ export default function TokenClaimModal(): ReactElement {
               )}
             </>
           )}
+          {!isClaimableNetwork && (
+            <p style={{ whiteSpace: "pre-line" }}>
+              <Trans i18nKey="disableRewardContent">
+                SDL is currently only deployed on Ethereum Mainnet and is not
```
ug02fast, line 124: It doesn't seem right to have the text content in the component and the translation file itself.

ug02fast (commented):
Looks fine other than having the same message in two places. Another thing is keeping formatting out of translation file `\n\n` is going to be cleaner. Using components to format the message will be cleaner.

src/components/TokenClaimModal.tsx
nkrulikovsky: I think, this is the ussage of <Trans> component. Because it is having <Link/> component, I have used <Trans/> component instead of useTranslation.

ug02fast: Good to know, thanks for the link.

ug02fast (approved):

hammeiam (approved):
Looks great!

nkrulikovsky:
> Looks fine other than having the same message in two places. Another thing is keeping formatting out of translation file `\n\n` is going to be cleaner. Using components to format the message will be cleaner.

Thanks to the function of `Trans` component (https://react.i18next.com/latest/trans-component#usage-with-simple-html-elements-like-less-than-br-greater-than-and-others-v10.4.0) we can use `br` component on the translation text directly but for some reason when we use br on our app
Screen Shot 2022-01-04 at 10 55 01 PM
 it shows me the following error.

ug02fast:
> > Looks fine other than having the same message in two places. Another thing is keeping formatting out of translation file `\n\n` is going to be cleaner. Using components to format the message will be cleaner.
> 
> Thanks to the function of `Trans` component (https://react.i18next.com/latest/trans-component#usage-with-simple-html-elements-like-less-than-br-greater-than-and-others-v10.4.0) we can use `br` component on the translation text directly but for some reason when we use br on our app Screen Shot 2022-01-04 at 10 55 01 PM it shows me the following error.
> 
> Absolutely to use simple html component on the translation text will be highly recommended @ug02fast

The app is crashing with this approach? Is it possible to format the text without using <br /> or any other html and use purely CSS?

nkrulikovsky:
> > > Looks fine other than having the same message in two places. Another thing is keeping formatting out of translation file `\n\n` is going to be cleaner. Using components to format the message will be cleaner.
> > 
> > 
> > Thanks to the function of `Trans` component (https://react.i18next.com/latest/trans-component#usage-with-simple-html-elements-like-less-than-br-greater-than-and-others-v10.4.0) we can use `br` component on the translation text directly but for some reason when we use br on our app Screen Shot 2022-01-04 at 10 55 01 PM it shows me the following error.
> > Absolutely to use simple html component on the translation text will be highly recommended @ug02fast
> 
> The app is crashing with this approach? Is it possible to format the text without using or any other html and use purely CSS?

To use other html component, we need to make the index as number on the translation file, I think  adding `\n\n` for breaking line is easier to use.

PR #758, opened 2022-01-03
Hide Gas Estimates

ug02fast:
Hiding gas estimates until solution to call cached BlockNative gas estimates exists.

image


hammeiam (approved):
I thought this should have effectively been hidden by having the gas price be 0, but it looks like that wasn't the case. I think it's good to keep this structure here so we can just uncomment it when we have gas prices again

Issue #760, opened 2022-01-04
Add back gas estimates


ug02fast:
Adding back gas estimates as gas estimate API no longer having CORS issue.

ug02fast:
not providing gas estimations anymore

PR #761, opened 2022-01-04
Reverted all instances of removed gas estimations

ug02fast:
Brought back gas estimations
image

https://github.com/saddle-finance/saddle-frontend/issues/760


src/utils/gas.ts
```diff
@@ -15,12 +15,12 @@ export function gasBNFromState(
   let gasPrice
   if (gasSelected === GasPrices.Custom) {
     gasPrice = gasCustom?.valueSafe
-  } else if (gasSelected === GasPrices.Fast) {
-    gasPrice = gasFast
+  } else if (gasSelected === GasPrices.Standard) {
+    gasPrice = gasStandard
```
ug02fast, line 19: Does the default to fast have to be adjusted here @hammeiam ?

src/state/user.ts
```diff
@@ -45,7 +45,7 @@ export const initialState: UserState = {
   userSwapAdvancedMode: false,
   userPoolAdvancedMode: false,
   userDarkMode: false,
-  gasPriceSelected: GasPrices.Standard,
+  gasPriceSelected: GasPrices.Fast,
```
ug02fast, line 48: These are a couple places I found throughout the codebase where defaults are set. Are there any others? @hammeiam

src/hooks/useApproveAndSwap.ts
```diff
@@ -89,7 +89,7 @@ export function useApproveAndSwap(): (
       } else {
         gasPrice = gasStandard
       }
-      gasPrice = parseUnits(gasPrice ? String(gasPrice) : "45", 9)
+      gasPrice = parseUnits(String(gasPrice), 9)
```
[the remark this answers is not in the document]
ug02fast: Yeah this number represents the normal gas price. We are tying to move away from hardcoding.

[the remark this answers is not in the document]
ug02fast: @nkrulikovsky Sorry I meant to say we are moving away from defaulting to standard Gas and towards fastGas. 45 used to represent standard gas but after EIP-1559 that isn't necessary.

src/components/ReviewDeposit.tsx
```diff
@@ -123,7 +123,7 @@ function ReviewDeposit({
             </span>
           </div>
         )}
-        {transactionData.txnGasCost?.valueUSD && (
+        {transactionData.txnGasCost?.valueUSD && shouldDisplayGas && (
```
hammeiam, line 126: is `valueUSD` a BigNumber? If so, `shouldDisplayGas` is probably redundant and we can just do `valueUSD.eq(Zero)`

ug02fast: good point. i did something similar using `gt` to avoid changing too much TS.

hammeiam:
Following a discussion, we're deciding to close this given that we don't yet have a good story around cross-chain gas estimation, and would prefer to push that responsibility to the user's wallet. We may come back to this once that's no longer the case.

PR #764, opened 2022-01-04
Account for paused status when getting swap pairs

hammeiam:
- get `paused` status for all pools via multicall (eventually to be replaced by graph)
- use paused status to remove tokens from swap UI
- use paused status to remove pools from swap route calculation


src/pages/Swap.tsx
```diff
@@ -145,6 +158,7 @@ function Swap(): ReactElement {
         }
       })
       .sort(sortTokenOptions)
+    console.log(allTokens.map(({ symbol }) => symbol).sort())
```
ug02fast, line 161:
```suggestion
```

ug02fast (changes requested):
looks good other than the console log

ug02fast (approved):

nkrulikovsky (approved):

PR #766, opened 2022-01-05
Migrate Confirm Bug

ug02fast:
Catch error on migrate confirm click.


src/components/PoolOverview.tsx
```diff
@@ -75,6 +75,12 @@ export default function PoolOverview({
     }),
   }
   const hasShare = !!userShareData?.usdBalance.gt("0")
+  console.log({
```
hammeiam, line 78: rm console log?

hammeiam (approved):

PR #770, opened 2022-01-06
Nikita/revamp theme

nkrulikovsky:
- Theme revamping.


src/providers/ThemeSettingsProvider.tsx
```diff
@@ -6,9 +6,10 @@ import React, {
   useEffect,
   useState,
 } from "react"
-import { darkTheme, lightTheme } from "../theme"
+import { darkTheme, lightTheme } from "../theme/theme"
```
ug02fast, line 9: No theme index? :'(

src/theme/components/AppbarTheme.ts
```diff
@@ -0,0 +1,15 @@
+import { Components, Theme } from "@mui/material"
+
+export default function AppBar(theme: Theme): Components {
+  console.log("background.default", theme.palette.background.default)
```
ug02fast, line 4: remove

src/theme/components/AppbarTheme.ts
```diff
@@ -0,0 +1,15 @@
+import { Components, Theme } from "@mui/material"
```
ug02fast, line 1: Rename file to AppBarTheme for correct casing?

src/theme/palette.ts
```diff
@@ -0,0 +1,66 @@
...
+  700: "#252525",
+}
+
+const lightPalette: PaletteOptions | undefined = {
+  mode: "light",
+  primary: {
+    main: saddleColors.seafoam,
+  },
+  secondary: {
+    main: saddleColors.sunburn,
```
ug02fast, line 44: does this approach give you autocompletion?

ug02fast (changes requested):
looks good just a couple small things

src/theme/palette.ts
nkrulikovsky: Yes.

src/providers/ThemeSettingsProvider.tsx
nkrulikovsky: Updated

src/theme/components/AppbarTheme.ts
nkrulikovsky: removed

src/theme/components/AppbarTheme.ts
```diff
@@ -1,7 +1,6 @@
 import { Components, Theme } from "@mui/material"
 
-export default function AppBar(theme: Theme): Components {
-  console.log("background.default", theme.palette.background.default)
+export default function AppBarTheme(theme: Theme): Components {
```
ug02fast, line 3: Do you mind renaming the file as well? `AppbarTheme -> AppBarTheme`

src/theme/components/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import BackdropTheme from "./BackdropTheme"
+import DialogTheme from "./DialogTheme"
+import TooltipTheme from "./TooltipTheme"
+import { merge } from "lodash"
+
+export default function componentsOverrides(theme: Theme): Components {
+  return merge(
+    AppBarTheme(theme),
+    DialogTheme(theme),
+    TooltipTheme,
```
hammeiam, line 12: Do these theme functions need to be called like the ones above them, or is it okay to merge in raw functions like this?

nkrulikovsky: These themes can merge in one function. Because there are much component in MUI so, I thought it would be more clean to split theme file for each components.

ug02fast: Why does `TooltipTheme` not have a parameter for theme? Will tooltip be the same style regardless of mode?

src/theme/palette.ts
```diff
@@ -0,0 +1,66 @@
...
+  sunset: "#D96A3B",
+  sand1: "#E3D899",
+  sand2: "#FAF3CE",
+  sand3: "#FDFDF8",
+  gold2: "#817F48",
+  cherry: "#FB5A5A",
+  cherry2: "#68282F",
+}
+
+const GREY = {
```
ug02fast, line 30: Changing this name to `GRAY_PALLETE` would help me understand it a bit more since it isn't exactly a single color.

nkrulikovsky: I recommend to use `GRAY_TONES` because there is `lightPalette and darkPalette name.

ug02fast (approved):

src/theme/components/index.ts
nkrulikovsky: It could have `theme` parameter, but it is not using theme for now. Actually `theme` parameter is used for using color and shadow. so `theme` parameter is optional and because of eslint, so I didn't give `theme` parameter for `TooltipTheme` for now. But we can add when we need.

hammeiam: My question was why are some component themes passed in as raw functions (`ToolipTheme`) rather than as executed functions (`TooltipTheme()`) to return the theme object?

src/theme/palette.ts
```diff
@@ -0,0 +1,66 @@
...
+    main: saddleColors.seafoam,
+  },
+  secondary: {
+    main: saddleColors.sunburn,
+  },
+  background: {
+    default: saddleColors.black,
+    paper: alpha(saddleColors.black, 0.7),
+  },
+  grey: GREY_TONES,
```
hammeiam, line 64: Are grey tones the same for light and dark themes? Does MUI handle inverting colors between modes automatically?

hammeiam (approved):
This looks great! I like how you split up the theme components into separate files.

src/theme/palette.ts
nkrulikovsky: Yes, it should be same, and MUI handles inverting colors automatically. Also we can give special colors for each component's theme as well.

src/theme/components/index.ts
nkrulikovsky: @hammeiam  You are correct.  component themes should be passed as executed functions.

PR #788, opened 2022-01-11
Update Account detail with MUI

nkrulikovsky:
- Add Button theme.
 - Replace AccountDetail dialog with Dialog component.
 - Update Palette (info color added)
 - Update Copy button with MUI button
 - Remove SCSS for account detail

## Context
https://www.figma.com/file/gqLcPTYXDlEsvYl4XDIV4i/?node-id=7065%3A51255

## Screenshot
Screen Shot 2022-01-11 at 6 27 12 PM
Screen Shot 2022-01-11 at 6 27 27 PM


src/components/DialogTitle.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onClose?: () => void
+}
+
+export default function DialogTitle({
+  children,
+  onClose,
+}: PropsWithChildren<unknown> & DialogTitleProps): ReactElement {
+  return (
+    <MuiDialogTitle>
+      <IconButton
```
hammeiam, line 16: Since this is absolute, could it overlap with `children` if they were long enough?

src/assets/icons/accountChange.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
nkrulikovsky: We can wrap SVG icon with <SvgIcon/> of MUI and create Icon component. (https://mui.com/components/icons/#svgicon)
In this case we need to scale for a 24x24px viewport svg Icon. 

With my experience it was easy to use SVG icon after import as ReactComponent.
For example
``` import {ReactComponent as SaddleIcon} from "../assest/saddle.svg" ```
After import like this we can give hight and width as Props on the Component (Like reward button)

In this case we need to give `currentColor` on the svg icon so it could have the wrapper's color.

For example:

```<svg width="52" height="52" viewBox="0 0 52 52" fill="currentColor" xmlns="http://www.w3.org/2000/svg">```
The advantage of this method is that we can use svg without make special component. But for the simple icon we often use, it might be good to wrap with <SvgIcon/>, I think.

File: src/components/DialogTitle.tsx (hunk 1 above)
Comment (nkrulikovsky) [reply src/components/DialogTitle.tsx:16]: Ah, Right.
On another branch I'll give close button on the <Dialog/> component since we don't use title on some component (like reward modal), so let me fix this on next branch, if you like. (we need to give higher zIndex for this button than dialog)

File: src/components/Copy.tsx [hunk 3]
```diff
@@ -8,40 +11,12 @@ export default function CopyHelper(props: {
...
-            d="M5.99998 2H12.6667C13.4 2 14 2.6 14 3.33333V10C14 10.7333 13.4 11.3333 12.6667 11.3333H11.2667V12.1333C11.2667 13.1274 10.4608 13.9333 9.46665 13.9333H3.86665C2.87254 13.9333 2.06665 13.1274 2.06665 12.1333V6.53333C2.06665 5.53922 2.87254 4.73333 3.86665 4.73333H4.66665V3.33333C4.66665 2.6 5.25998 2 5.99998 2ZM4.66665 5.93333H3.86665C3.53528 5.93333 3.26665 6.20196 3.26665 6.53333V12.1333C3.26665 12.4647 3.53528 12.7333 3.86665 12.7333H9.46665C9.79802 12.7333 10.0667 12.4647 10.0667 12.1333V11.3333H5.99998C5.25998 11.3333 4.66665 10.7333 4.66665 10V5.93333ZM5.99998 10H12.6667V3.33333H5.99998V10Z"
-          />
-        </svg>
-      )}
-      {isCopied ? "" : props.children}
-    </button>
+    <Button
+      color="info"
+      onClick={() => setCopied(props.toCopy)}
+      startIcon={isCopied ? <CheckIcon /> : <ContentCopyIcon />}
```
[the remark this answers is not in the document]
nkrulikovsky: 🙏🏻

cypress/integration/accountDetail.test.ts
```diff
@@ -7,10 +7,10 @@ context("Account detail test", () => {
...
 
   it("renders account detail after click account avatar", () => {
-    cy.get(".walletStatus").should("exist")
-    cy.get(".walletStatus>button").click()
-    cy.get(".accountDetail").should("exist")
-    cy.get(".buttonGroup>button").contains("Change Account").click()
-    cy.get(".modal").should("exist")
+    cy.get("[data-testid=walletStatusContainer]").should("exist")
+    cy.get("[data-testid=accountDetailButton]").should("exist").click()
+    cy.get("[data-testid=accountDetailContainer]").should("exist")
```
ug02fast, line 12: this is correct

src/components/AccountDetails.tsx
```diff
@@ -27,87 +41,87 @@ export default function AccountDetail({ openOptions }: Props): ReactElement {
...
+          <Box display="flex" alignItems="center">
+            {account && <Copy toCopy={account}>{t("copyAddress")}</Copy>}
+          </Box>
+          <Box display="flex" alignItems="center">
+            <Button
               onClick={() => {
                 openOptions()
               }}
+              startIcon={<ChangeIcon />}
+              data-testid="changeAccount"
```
ug02fast, line 107: This is a step in the right direction. If you name it something like `changeAccountBtn` it will avoid clashing if in the future there is another place in the app where there is a change account function.

src/components/ConnectWallet.tsx
```diff
@@ -16,7 +16,7 @@ function ConnectWallet({ onClose }: Props): ReactElement {
   const { activate } = useWeb3React()
 
   return (
-    <div className="connectWallet">
+    <div className="connectWallet" data-testid="connectWallet">
```
ug02fast, line 19: Step in the right direction. I'd make it a little more descriptive i.e. `connectWalletContainer`

src/components/Copy.tsx
```diff
@@ -8,40 +11,11 @@ export default function CopyHelper(props: {
...
-          />
-        </svg>
-      )}
-      {isCopied ? "" : props.children}
-    </button>
+    <Button
+      onClick={() => setCopied(props.toCopy)}
+      startIcon={isCopied ? <CheckIcon /> : <ContentCopyIcon />}
+    >
+      {isCopied ? "Copied!" : props.children}
```
ug02fast, line 18: Are we translating here or is it not necessary?

ug02fast (approved):

src/components/Copy.tsx
nkrulikovsky: Yes we need.

src/components/AccountDetails.tsx
```diff
@@ -27,87 +41,87 @@ export default function AccountDetail({ openOptions }: Props): ReactElement {
   )
 
   const connectorName = find(SUPPORTED_WALLETS, ["connector", connector])?.name
+  const theme = useTheme()
+  theme.palette.background
```
hammeiam, line 45: rm

src/theme/components/DialogTheme.ts
```diff
@@ -3,11 +3,19 @@ import { Components, Theme } from "@mui/material"
...
+        BackdropProps: {
+          timeout: 500,
+        },
+      },
       styleOverrides: {
         paper: {
           borderRadius: 10,
-          border: `1px solid ${theme.palette.grey[500]}`,
-          backgroundColor: theme.palette.background.default,
+          // TDDO: remove this hardcoded color
```
hammeiam, line 14: rm?

src/theme/components/PaperTheme.ts
```diff
@@ -1,11 +1,14 @@
...
-export default function PaperTheme(): Components {
+export default function PaperTheme(theme: Theme): Components {
   return {
     MuiPaper: {
       styleOverrides: {
         root: {
-          border: "1px solid #000000",
+          border:
+            theme.palette.mode === "light"
+              ? theme.palette.other?.divider
```
hammeiam, line 10: should be `1px solid ${theme.palette.other?.divider}`

src/theme/palette.ts
```diff
@@ -34,32 +23,69 @@ const GREY_TONES = {
   500: "#404040",
   700: "#252525",
 }
-
 const lightPalette: PaletteOptions | undefined = {
   mode: "light",
   primary: {
-    main: saddleColors.seafoam,
+    main: "#06D7D7",
```
hammeiam, line 29: Why did we remove `saddleColors`?

hammeiam (approved):
I approve to unblock, however there are a couple comments to address.

src/theme/palette.ts
nkrulikovsky: After giving color on the theme we will use these color over all the theme. So, actually we only need this color on the theme file. And it would be not convenient to give name on saddle color , also on the editor, after installing extension, we can see color on the hex or rgb color.
So, I removed saddleColor.

Also if we need another color name it would be good to give on the palette theme after declare types. Then we can access to this color from anywhere of our app.

src/theme/components/PaperTheme.ts
nkrulikovsky: Exactly 😊

src/components/AccountDetails.tsx
nkrulikovsky: removed

Issue #795, opened 2022-01-21
[Bug] Support ENS on non-ENS chains


hammeiam:
web3 provider will throw an error when trying to look up ENS on unsupported chains (I believe only mainnet supports)

Either:
Follow this guide https://medium.com/the-ethereum-name-service/how-to-support-ens-for-multi-chain-dapps-b0a7ff043d77
Or: 
Disable ens on non-mainnet chains

PR #802, opened 2022-01-24
Support ENS name on non-mainnet

nkrulikovsky:
https://github.com/saddle-finance/saddle-frontend/issues/795


src/hooks/useENS.ts
```diff
@@ -1,23 +1,41 @@
...
+        chainId === ChainId.MAINNET
+      ) {
         const provider = new ethers.providers.Web3Provider(library.provider)
         const ensName = await provider.lookupAddress(address)
+        console.log("ens name on net", ensName)
         if (ensName) setENSName(ensName)
+      } else if (
+        address &&
+        library &&
+        ethers.utils.isAddress(address) &&
```
ug02fast, line 27: not necessary but putting this logic into a const could clean up a couple lines.

src/hooks/useENS.ts
```diff
@@ -1,23 +1,41 @@
...
+        chainId !== ChainId.MAINNET
+      ) {
+        // Get ens name from main chain when network is non-mainnet
+        const ensNameOnMainnet = await ethers
+          .getDefaultProvider()
```
ug02fast, line 32: What is the difference between .getDefaultProvider() and Web3Provider.

ug02fast (approved):
looks good just left a couple comments

src/hooks/useENS.ts
```diff
@@ -1,23 +1,40 @@
...
```
hammeiam, line 31: Can we use REACT_APP_NETWORK_URL as an argument here? My concern is that `getDefaultProvider` basically spams free endpoints, meaning that it creates a ton of network requests and will eventually return 429s or something like that. Here's was the network looks like when moving to another chain: 

Screen Shot 2022-01-27 at 7 15 54 PM

nkrulikovsky: You are right.

ug02fast (approved):

hammeiam (approved):
Works well, thank you Nikita.eth 🤠

PR #803, opened 2022-01-25
Remove V2 terminology from pools on the UI.

ug02fast:
Removing v2 terminology from the pool names in favor of distinguishing v1/v2
pools with the "outdated" tag.


ug02fast:
@hammeiam hacky workaround but provides quick solution to renaming without breaking the UI. could be revisited after changing approach to pulling pool data. thoughts?

ug02fast:
Need to revisit when it's important. Closing for now

PR #823, opened 2022-02-03
Toast Notification Rework

ug02fast:
image

The new Toast API has two functions that can trigger a Toast notification.

One is ideal for promise-based notifications the other is for strictly synchronous executions. The synchronous-based one is better for triggering a toast when an error occurs for a transaction that we are not explicitly showing a notification flow for. The promise-based toast is ideal for most scenarios and was used in place of the previous `notifyHandler` calls.

The new API allows for more extensibility than the previous solution which allows us to tune the UX to a much larger degree than before.

The new toast library comes with many UX improvements that can be adjusted over time depending on user needs.


src/utils/checkAndApproveTokenForTrade.ts
```diff
@@ -50,16 +44,14 @@ export default async function checkAndApproveTokenForTrade(
...
-      const cleanupOnStart = callbacks.onTransactionStart?.()
       const approvalTransaction = await srcTokenContract.approve(
         spenderAddress,
         amount,
       )
-      // Add notification
-      notifyHandler(approvalTransaction.hash, "tokenApproval")
-      const confirmedTransaction = await approvalTransaction.wait()
-      cleanupOnStart?.()
-      callbacks.onTransactionSuccess?.(confirmedTransaction)
```
ug02fast, line 62: @hammeiam fyi: deleting these as they aren't being used

src/components/TokenClaimDialog.tsx
```diff
@@ -163,18 +163,17 @@ export default function TokenClaimDialog({
               </p>
             )}
             {allPoolsWithRewards.map((pool, i, arr) => (
-              <>
+              <React.Fragment key={pool.name}>
```
ug02fast, line 166: React key bug fix

src/hooks/useRewardsHelpers.ts
```diff
@@ -51,16 +51,18 @@ export function useRewardsHelpers(poolName: PoolName): {
           BigNumber.from(1),
         )
         const txn = await rewardsContract.deposit(poolPid, amount, account)
-        notifyHandler(txn?.hash, "deposit")
-        await txn.wait()
+        await enqueuePromiseToast(txn.wait(), "deposit", { poolName })
+        // notifyHandler(txn?.hash, "deposit")
+        // await txn.wait()
```
ug02fast, line 56: this isn't actually doing anything

sandrasong:
Tested in local testnet, works smooth. The style looks good to me. 

Just one small change if it's possible. Could you use stake & unstake for staking / unstaking txn? It says deposit & withdraw now. 
Screen Shot 2022-03-02 at 6 59 55 PM
png">

ug02fast:
> 

Yes, good catch @sandrasong

src/components/Toastify.tsx
```diff
@@ -0,0 +1,114 @@
...
+        return "Swap Initiated"
+      case "withdraw":
+        return "Withdraw Initiated"
+      case "claim":
+        return "Claim Initiated"
+      case "migrate":
+        return "Migrate Initiated"
+      case "stake":
+        return "Stake Initiated"
+      case "unstake":
```
hammeiam, line 41: what do you think of making these constants eg `TXN_TYPES.STAKE` to prevent me from typing `wtihdraw` and showing the wrong UI?

ug02fast: since these are typed, tsc will prevent misspelling of potential arguments you can pass. it will also auto complete as soon as you start typing out the fn i.e:

image
image

src/components/Toastify.tsx
```diff
@@ -0,0 +1,105 @@
...
+              display: "flex",
+              justifyContent: "space-around",
+              alignItems: "center",
+            }}
+          >
+            <>
+              {renderSuccessContentBasedOnType(type)}
+              <Link
+                // @ts-ignore
+                href={getEtherscanLink(data?.transactionHash, "tx")}
```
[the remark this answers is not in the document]
ug02fast: good catch

src/components/PendingSwapModal.tsx
```diff
@@ -20,13 +21,13 @@ import { Zero } from "@ethersproject/constants"
...
 import { formatUnits } from "@ethersproject/units"
 import { gasBNFromState } from "../utils/gas"
-import { notifyHandler } from "../utils/notifyHandler"
 import styles from "./PendingSwapModal.module.scss"
 import { subtractSlippage } from "../utils/slippage"
 import { useBridgeContract } from "../hooks/useContract"
 import { useEffect } from "react"
 import { useSelector } from "react-redux"
 import { useTranslation } from "react-i18next"
+import { useWeb3React } from "@web3-react/core"
```
hammeiam, line 30: should be using `import { useActiveWeb3React } from "../hooks"`

src/utils/getEtherscanLink.ts
```diff
@@ -1,3 +1,31 @@
+import { ChainId } from "../constants"
+
+export function getMultichainScanLink(
```
hammeiam, line 3: Awesome!

hammeiam (approved):

PR #837, opened 2022-02-14
Update swap button with MUI

nkrulikovsky:
Update Swap button with MUI

### DIFF
--- a/src/components/SwapPage.tsx
+++ b/src/components/SwapPage.tsx
@@ -1,5 +1,6 @@
 import "./SwapPage.scss"
 
+import { Box, Button } from "@mui/material"
 import React, { ReactElement, useMemo, useState } from "react"
 import { SWAP_TYPES, getIsVirtualSwap } from "../constants"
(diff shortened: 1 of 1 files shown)


src/components/SwapPage.tsx
```diff
@@ -295,9 +294,12 @@ const SwapPage = (props: Props): ReactElement => {
             )
           })}
         </div>
-        <Box mt={3}>
+        <Box mt={3} width="100%">
```
ug02fast, line 297: Is it possible to accomplish this utilizing flexbox and not explicitly setting width to `100%`? I can imagine it can be a bit of a process with the migration to mUI as well.

nkrulikovsky: Right. We don't need to give `100%` if we will use flexbox. Let me update this accordingly removing `SCSS` file of `Swap` page.

ug02fast (approved):

hammeiam (approved):

PR #870, opened 2022-02-28
Tooltip update

nkrulikovsky:
## What does this PR do?
- Replace tooltip on the `PoolInfoCard` and `PoolOverview`
- Remove Tooltip.scss and `ToolTip` component

## Preview
https://bafybeibzdj6cwgz76smapimtcu4pbck5slqzc4f6zcm2sog2iifs6f3lwi.on.fleek.co/


hammeiam (approved):

src/components/PoolOverview.tsx
```diff
@@ -86,9 +85,9 @@ export default function PoolOverview({
       <div className="left">
         <div className="titleAndTag">
           {isMetapool ? (
-            <ToolTip content={t("metapool")}>
+            <Tooltip title={<React.Fragment>{t("metapool")}</React.Fragment>}>
```
ug02fast, line 88: This prop won't take just strings 👀 ?

ug02fast (approved):

src/components/PoolOverview.tsx
nkrulikovsky: We can use `string` and `React component` for this props.
FYI : https://mui.com/components/tooltips/#customization

PR #877, opened 2022-03-03
update confirm transaction

nkrulikovsky:


ug02fast (approved):
looks good. is it possible to extract the svg into a file and import it instead of explicitly showing it's full form here?

nkrulikovsky:
Importing logo file will make code more clean.

hammeiam (approved):

PR #878, opened 2022-03-03
Update PoolOverview with MUI

nkrulikovsky:


src/components/PoolOverview.tsx
```diff
@@ -75,141 +86,197 @@ export default function PoolOverview({
...
+                  >
                     <Tooltip
                       title={
                         <React.Fragment>
                           {symbol.replaceAll("/", "\n")}
                         </React.Fragment>
                       }
                     >
-                      <React.Fragment>Reward APR</React.Fragment>
+                      <React.Fragment>Reward APR: </React.Fragment>
```
ug02fast, line 229: Could you remove the Fragment and wrap the tooltip around the `Typography` here?

ug02fast (changes requested):
looks fine other than moving tooltip around Typography component

src/components/PoolOverview.tsx
nkrulikovsky: That way is more obvious even though I have already gave font for tooltip on `theme.ts`!

src/components/PoolOverview.tsx
```diff
@@ -75,141 +87,191 @@ export default function PoolOverview({
...
-        </div>
-      </div>
+      <Grid container alignItems="center" spacing={1}>
+        <Grid item xs={12} lg={3}>
+          <Box>
+            <Box
+              display="flex"
+              color={disableText ? theme.palette.text.disabled : undefined}
+            >
+              <Tooltip title={isMetapool ? <div>{t("metapool")}</div> : ""}>
```
nkrulikovsky, line 126: FYI, if title of `Tooltip` is empty("") then tooltip don't appear.
https://mui.com/api/tooltip/#props

src/components/PoolOverview.tsx
nkrulikovsky: Fixed!

package-lock.json
```diff
@@ -31,7 +31,7 @@
         "@web3-react/injected-connector": "^6.0.7",
         "@web3-react/network-connector": "^6.1.9",
         "@web3-react/types": "^6.0.7",
-        "@web3-react/walletconnect-connector": "^6.2.10",
+        "@web3-react/walletconnect-connector": "^6.2.13",
```
hammeiam, line 34: your `package-lock` should not be changing unless your `package.json` is. Please use `npm ci` instead of `npm i`/`npm install` unless you're adding a new dependency.

src/components/PoolOverview.tsx
```diff
@@ -75,141 +87,191 @@ export default function PoolOverview({
...
+                {t("migrate")}
+              </Button>
+            ) : (
+              <Button
+                variant="contained"
+                color={isOutdated || shouldMigrate ? "secondary" : "primary"}
+                fullWidth
+                size="large"
                 disabled={poolData?.isPaused || isOutdated}
+                onClick={() => history.push(`${poolRoute}/deposit`)}
```
hammeiam, line 253: What is the logic of using a Button with `history.push` rather than just using `Link`?

nkrulikovsky: Because we removed css, even though we disable `Button`, it navigate with the help of `Link`. Of course, we can customize link with `styled` but to reduce code I use `hisotry.push` and removed `Link`.

package-lock.json
nkrulikovsky: Does it work when there is conflict on `package-lock.json` because I am knowing that `npm ci` command can only install with an existing package-lock.json.

hammeiam:
Please remove `package-lock.json` changes from this PR unless you're also changing `package.json`

nkrulikovsky:
@hammeiam 
Removed

hammeiam (approved):

src/components/PoolOverview.tsx
```diff
@@ -9,18 +22,16 @@ import {
...
   formatBNToShortString,
   formatBNToString,
 } from "../utils"
 
-import Button from "./Button"
-import { Link } from "react-router-dom"
 import { Zero } from "@ethersproject/constants"
-import classNames from "classnames"
 import logo from "../assets/icons/logo.svg"
+import { useNavigate } from "react-router-dom"
```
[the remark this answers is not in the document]
nkrulikovsky: OK

ug02fast (approved):

PR #880, opened 2022-03-08
update ReviewDeposit and remove ReviewDeposit.scss

nkrulikovsky:
## What does this PR do?

Update the `ReviewDeposit.tsx` with MUI

## Screenshot

image

image


nkrulikovsky:
NOTE: `HightImpact` section will be update on another branch.

src/components/ReviewDeposit.tsx
```diff
@@ -49,80 +55,99 @@ function ReviewDeposit({
     transactionDeadlineCustom,
   )
   const shouldDisplayGas = !!gasStandard
+
+  const DepositeInfoItem = styled(Box)(({ theme }) => ({
```
ug02fast, line 59:
```suggestion
  const DepositInfoItem = styled(Box)(({ theme }) => ({
```

src/components/ReviewDeposit.tsx
```diff
@@ -49,80 +55,99 @@ function ReviewDeposit({
...
+    display: "flex",
+    minWidth: "100%",
+    marginBottom: theme.spacing(2),
+    "&>:first-child": {
```
ug02fast, line 63: Can this flex solution be done on the component level and outside of explicit styling? Why is it that only the first and last childs are problematic with styles?

ug02fast (changes requested):

src/components/ReviewDeposit.tsx
nkrulikovsky: There are several `DepositInfo`, so. i made this `component` to avoid to giving props repeatly.
And, by giving property first and last component, It doesn't need to props on the child component in `DepositInfoItem`.

src/components/ReviewDeposit.tsx
```diff
@@ -24,6 +30,19 @@ interface Props {
   transactionData: DepositTransaction
 }
 
+const DepositeInfoItem = styled(Box)(({ theme }) => ({
```
ug02fast, line 33: Could you also fix spelling here and in all instances where `Deposite` is used?

ug02fast (approved):

PR #883, opened 2022-03-10
fix: toast theme according the mui theme mode

nkrulikovsky:
Change `Toast theme` according to the `MUI` theme.

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -1,7 +1,6 @@
 import "../styles/global.scss"
 import "react-toastify/dist/ReactToastify.css"
 
-import { AppDispatch, AppState } from "../state"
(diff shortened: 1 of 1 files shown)


src/pages/App.tsx
```diff
@@ -12,8 +11,9 @@ import React, {
   useMemo,
 } from "react"
 import { Route, Routes } from "react-router-dom"
-import { useDispatch, useSelector } from "react-redux"
+import { styled, useTheme } from "@mui/material"
```
ug02fast, line 14: do you need styled?

nkrulikovsky: Yeah, that was used for the background of the app.

PR #884, opened 2022-03-10
Cypress upgrade. withdrawal e2e adjustments.

ug02fast:
~~Removed all `wait()` calls.~~ Due to strange latency in container instance added back wait() calls.

Added a cypress config that hides all xhr/fetch requests in logs which was providing no value.

Made some changes that tell Cypress to wait for UI changes to occur before moving to the next assertion.

Added a GH Action that uploads video and screenshot artifacts if a test fails.

Upgraded: `@cypress/code-coverage` and `cypress` to latest versions.

closes: https://github.com/saddle-finance/saddle-frontend/issues/900


cypress/integration/withdrawal.test.ts
```diff
@@ -19,152 +14,88 @@ context("Withdrawal Flow", () => {
...
+          const prevVal = $value.text()
+          cy.get("#tokenInput input").then(($inputs) => {
+            cy.wrap($inputs).each(($input) => {
+              cy.wrap($input).type("2")
             })
+            cy.get("button").contains("Withdraw").click()
+            cy.get("button").contains("Confirm Withdraw").click()
+            cy.get("[data-testid=tokenValue]")
+              .first()
+              .should("not.have.text", prevVal)
```
hammeiam, line 96: this would pass if the value is null-ish, right? Any reason to make the test less precise in this way?

ug02fast: This assertion will wait for the `tokenValue <p>` in the UI to not have the `prevVal` from before the `Confirm Withdraw` was clicked. I.E It's watching the value initially at 96.01 then passes the assertion only after the state change implies the withdrawal is complete, which would show 94.01. It should never get to a null-ish value here.

hammeiam: I guess I'm saying if the state changes from 94 to NaN, this passes, correct?

ug02fast: Theoretically yes but it's not any different from the last solution in that respect. The important thing with this check is to make sure something on the UI changed indicating that the withdrawal transfer transaction occurred and returned a new value.

hammeiam (approved):

hammeiam (approved):

PR #886, opened 2022-03-14
Update review migration and create alert theme

nkrulikovsky:
## What does this PR do?
Update review migration and create `Alert` theme and remove warning component from the dapp.

## Screenshot

image
Screen Shot 2022-03-14 at 3 53 46 PM
image


ug02fast (changes requested):

src/components/ReviewMigration.tsx
```diff
@@ -45,55 +50,73 @@ function ReviewMigration({
...
+    <React.Fragment>
+      <DialogTitle variant="h1" onClose={onClose}>
+        {t("reviewMigration")}
+      </DialogTitle>
+      <DialogContent>
+        <Alert variant="filled" icon={false} severity="warning">
+          {t("migrationExplain")}
+        </Alert>
+        <Box my={3}>
+          <Box>
```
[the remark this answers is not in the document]
nkrulikovsky: You are right!

src/components/ReviewMigration.tsx
```diff
@@ -45,55 +50,73 @@ function ReviewMigration({
...
-            </div>
-          )}
-          {/* TODO: Create a light API to expose the cached BlockNative gas estimates. */}
-          {/* {gasValueUSD && (
+    <React.Fragment>
+      <DialogTitle variant="h1" onClose={onClose}>
+        {t("reviewMigration")}
+      </DialogTitle>
+      <DialogContent>
+        <Alert variant="filled" icon={false} severity="warning">
```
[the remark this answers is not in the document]
nkrulikovsky: we can give `false` when we don't need icon on the Alert.

FYI, https://mui.com/api/alert/
image

src/components/ReviewMigration.tsx
```diff
@@ -45,55 +50,67 @@ function ReviewMigration({
...
+          {t("migrationExplain")}
+        </Alert>
+        <Box my={3}>
+          <Box display="flex">
+            <Typography component="span">{t("migrationAmount")}</Typography>
```
ug02fast, line 63: What is the reason for explicitly making passing `span` into this prop? Does it need it?

nkrulikovsky: Oh, the component props render `Typography` as that component. So the result will be `span`.
Because this component was `span` so I gave this props.

but as you said because it was wrapped with flex box, so that props don't mandatory.

ug02fast: Are you saying by default Typography renders as a `span`?

nkrulikovsky: Agreed!

hammeiam (approved):

ug02fast (approved):

PR #887, opened 2022-03-14
Update review swap with MUI

nkrulikovsky:


ug02fast (approved):

src/components/ReviewSwap.tsx
```diff
@@ -103,21 +114,21 @@ function ReviewSwap({ onClose, onConfirm, data }: Props): ReactElement {
...
               {formatBNToString(data.exchangeRateInfo.exchangeRate, 18, 6)}
-            </span>
-          </div>
-          <div className="row">
-            <span className="title">{t("gas")}</span>
-            <span className="value floatRight">
+            </Typography>
+          </Box>
+          <Box display="flex">
+            <Typography component="span">{t("gas")}</Typography>
```
hammeiam, line 122: We should be hiding gas - we don't currently have a good way of estimating this value and instead leave it up to the user's wallet

hammeiam (approved):
one change requested

PR #889, opened 2022-03-15
Reverting FE Tune-Up dep upgrades

ug02fast:
Upgrading all deps at once breaks wallet-connect and Tally wallet. Fixing these involves many changes to the @web3-react library which creates a chain reaction of work that is too much to do in one go.

Solution to making these upgrades is upgrading deps individually instead of in bulk.


hammeiam (approved):
Approved! What's something we can improve in our CR's to reduce reverts in the future? Testing wallets manually?

ug02fast:
@hammeiam 

> Approved! What's something we can improve in our CR's to reduce reverts in the future? Testing wallets manually?

E2E tests that at least asserts that the initial wallet plugin modal renders. Most of the time that is the first indicator that a wallet is broken.

src/components/TopMenu.tsx
```diff
@@ -111,7 +111,7 @@ function TopMenu(): ReactElement {
             <NetworkDisplay onClick={handleSettingMenu} />
             <IconButton
               onClick={handleSettingMenu}
-              data-testid="settingsMenuBtn"
```
ug02fast, line 114: fix for wrong casing

cypress/support/index.ts
```diff
@@ -19,10 +19,4 @@ import "cypress-react-selector"
 import "@cypress/code-coverage/support"
 
 // Alternatively you can use CommonJS syntax:
-// require("./commands")
-
-Cypress.on("uncaught:exception", () => {
```
ug02fast, line 24: fix for not failing on uncaught

PR #918, opened 2022-03-22
Config for recording failed E2E fix/workaround.

ug02fast:
This is the best solution when it comes to recording failed e2e tests that allows us to have videos to debug with, without having to pay.


cypress.json
```diff
@@ -10,5 +10,6 @@
     }
   },
   "defaultCommandTimeout": 30000,
-  "requestTimeout": 30000
+  "requestTimeout": 30000,
+  "videoCompression": false
```
hammeiam, line 14: Why no compression?

hammeiam (approved):

cypress.json
ug02fast: Good call. I was under the assumption that Cypress was going to record and compress a video for every test which is not actually the case.

Issue #893, opened 2022-03-15
Scrollbar on Windows changes colors when dropdown is clicked


ug02fast:
![brave_bDBuAfq0nS](https://user-images.githubusercontent.com/5507707/158454653-ba8f3a9d-9834-4d34-a827-bb410fb67133.gif)

PR #926, opened 2022-03-25
Fix scrollbar color issue appearing when Menu open

nkrulikovsky:
## What does this PR do?

close: https://github.com/saddle-finance/saddle-frontend/issues/893

https://user-images.githubusercontent.com/75422800/160211363-544627b5-ab4f-4c1b-9b6e-86ee7435ae2e.mov


ug02fast:
Can you make it so the overflow bar doesn't even render since there is not vertical scrolling on the page?

nkrulikovsky:
Scrollbar was added to preventing the shake of page with and without scrollbar. So, I made vertical scrollbar visible and gave the margin left of `margin-right: calc(-1 * (100vw - 100%));`

Yes, we can hide scrollbar when there is no scroll like swap page as well because all component updated with MUI and it doesn't overflow horizontally.

hammeiam (approved):

Issue #924, opened 2022-03-25
Faster integration tests


hammeiam:
deposit needs to run first, swap & withdraw need to run after it, but the rest of the tests can be run in parallel. Add some config to cypress to run - to keep it simple, just group the stateful tests together and the non-stateful tests and run each as its own test group https://docs.cypress.io/guides/guides/parallelization

PR #927, opened 2022-03-26
Separate Cypress tests into separate groups

nguyenaiden:
Description: Currently, Cypress integration tests are slow because they're all running serially. The solution is to separate tests into groups - stateful and stateless, with the latter one running in parallel:: 
- stateful tests (deposit, swap, withdraw) 
- stateless tests (accountDetails, advancedOptions, risk, and topMenu)

This change also requires ci.yml to be updated to accommodate the changes. 
[Related ticket]( https://github.com/saddle-finance/saddle-frontend/issues/924)


.github/workflows/ci.yml
```diff
@@ -94,9 +94,17 @@ jobs:
       # fix for forcing git to use https when pulling deps
       - run: 'git config --global --replace-all url."https://github.com/".insteadOf ssh://git@github.com/'
       - run: npm ci
-      - run: npm run cy:coverage
+      - name: stateful tests
```
hammeiam, line 97: these steps will run back to back - it may be better to define a separate job for these so they can each have their own machine. If we do this then we won't necessarily need to use the `parallel` flag since we're basically manually parallelizing

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "saddle-frontend",
```
hammeiam, line 8: 👀  not really sure why this happens but please rm from PR

.github/workflows/ci.yml
```diff
@@ -76,7 +76,46 @@ jobs:
...
+  coverage-cypress-stateless:
+    runs-on: ubuntu-latest
+    services:
+      saddle-contract:
+        image: saddlefinance/contracts:latest
+        ports:
+          - 8545:8545
+    strategy:
+      fail-fast: false
+      matrix:
```
hammeiam, line 88: we can probably just remove `matrix` altogether if it's being run on one server

PR #928, opened 2022-03-28
Replace KEEP with T rewards

hammeiam:
Followed the KEEP+Nu merge, KEEP incentives are being migrated to T (Threshold network token). Read for more info https://forum.keep.network/t/repurpose-saddle-tbtc-pool-liquidity-incentives-and-move-incentives-to-t/404

Screen Shot 2022-03-27 at 11 50 49 PM
Screen Shot 2022-03-27 at 11 50 38 PM


src/react-app-env.d.ts
```diff
@@ -33,3 +33,16 @@ interface Window {
...
+  ? (keyof T)[]
+  : T extends number
+  ? []
+  : T extends Array<any> | string
+  ? string[]
+  : never
+
+interface ObjectConstructor {
+  keys<T>(o: T): ObjectKeys<T>
+}
```
nkrulikovsky, line 48: Where does we need this type?

src/utils/thirdPartyIntegrations.ts
```diff
@@ -317,8 +276,47 @@ async function getSperaxData(
...
+    await Promise.all([
+      lpTokenContract.balanceOf(MINICHEF_CONTRACT_ADDRESSES[chainId]),
+      rewardsContract.rewardPerSecond(),
+      rewardsContract.userInfo(accountId || AddressZero),
+    ])
+
+  const spaPerYear = thresholdRewardsPerSecond.mul(3600 * 24 * 365) // 1e18
+  const spaPerYearUSD = spaPerYear.mul(parseUnits(thresholdPrice.toFixed(3), 3)) // 1e18 + 3 = 1e21
+  const totalDepositedUSD = totalDeposited.mul(lpTokenPrice) // 1e18 + 1e18 = 1e36
+  const spaApr = shiftBNDecimals(spaPerYearUSD, 33).div(totalDepositedUSD) // (1e21 + 1e33 = 1e54) / 1e36 = 1e18
```
penandlim, line 320: spa -> threshold

src/react-app-env.d.ts
hammeiam: This type is used wherever we use `Object.keys`. By default, the type of `keys` is `string[]`, but it should more literally be `keyof typeof myObject`. Doing `Object.keys(myObject).map(key => myObject[key])` may throw a TS warning since you can't index `myObject` with `string`, and this interface resolves that.

ug02fast (approved):

ug02fast (approved):

src/react-app-env.d.ts
nkrulikovsky: Makes a sense!

nkrulikovsky (approved):

PR #931, opened 2022-03-29
Multicall3 tryEach Refactor

ug02fast:
Refactored all `.all()` multicalls to `.tryEach()`.

Updated typings.

Updated multicall3 address for hardhat.

### DIFF
--- a/src/hooks/usePoolsStatuses.ts
+++ b/src/hooks/usePoolsStatuses.ts
@@ -56,8 +56,14 @@ export default function usePoolStatuses(): PoolStatuses {
         })
         .map((c) => c.paused())
       try {
-        const tvls = await ethcallProvider.all(supplyCalls)
-        const pausedStatuses = await ethcallProvider.all(pausedCalls)
+        const tvls = await ethcallProvider.tryEach(
+          supplyCalls,
+          supplyCalls.map(() => true),
+        )
+        const pausedStatuses = await ethcallProvider.tryEach(
+          pausedCalls,
+          pausedCalls.map(() => true),
+        )
         const tvlsUSD = pools.map((pool, i) => {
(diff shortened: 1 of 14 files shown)


src/types/ethcall.d.ts
```diff
@@ -72,7 +75,12 @@ export class MulticallProvider extends Provider {
     D extends MulticallCall,
   >(
     calls: readonly [A, B, C, D],
+    canFail: boolean[],
     block?: number,
   ): Promise<[A["outputs"], B["outputs"], C["outputs"], D["outputs"]]>
-  tryEach<O>(calls: MulticallCall<I, O>[], block?: number): Promise<O[]> // fallthrough to array type
+  tryEach<O>(
```
hammeiam, line 81: tbh I'm not sure how to type this off the top of my head, but now any call result may also be null if `canFail = true`.

src/hooks/usePoolsStatuses.ts
```diff
@@ -56,8 +56,14 @@ export default function usePoolStatuses(): PoolStatuses {
         })
         .map((c) => c.paused())
       try {
-        const tvls = await ethcallProvider.all(supplyCalls)
-        const pausedStatuses = await ethcallProvider.all(pausedCalls)
+        const tvls = await ethcallProvider.tryEach(
+          supplyCalls,
+          supplyCalls.map(() => true),
```
hammeiam, line 61: for safety / backwards compatibility's sake, lets mark each of these tryEach calls as `canFail=false`. As it currently is, one of these calls returning `null` will break stuff since our type doesn't account for possible failures. We should still get the performance boost from using the new contracts.

hammeiam (approved):
change all `canFail` args to `false`, then you're good to land

PR #937, opened 2022-03-30
Permissionless Pool creation integrated with contract.

ug02fast:
Complete Create Permissionless Pool page now calling to contract methods appropriately.

Some other takeaways are that this is the first component utilizing the `poolRegistry` hook.

The create page is also able to extend to support custom metapool basepools if we choose so in the future. The token input validates a valid token address has been inputted.

image

image
closes: #908


package-lock.json
```diff
@@ -10552,7 +10552,7 @@
       "version": "0.0.8",
       "resolved": "https://registry.npmjs.org/ansi-html-community/-/ansi-html-community-0.0.8.tgz",
       "integrity": "sha512-1APHAyr3+PCamwNw3bXCPp4HFLONZt/yIH0sZp0/469KWNTEy+qN5jQ3GVX6DMZ1UXAi34yVwtTeaG/HpBuuzw==",
-      "devOptional": true,
+      "dev": true,
```
hammeiam, line 10555: why is package-lock changing if `package.json` is unchanged?

src/constants/index.ts
```diff
@@ -121,8 +121,15 @@ export class Token {
 
 export const BLOCK_TIME = 13000 // ms
 
+// TODO: Update with mainnet and ropsten addresses
+export const PERMISSIONLESS_DEPLOYER_CONTRACT_ADDRESSES = buildAddresses({
```
hammeiam, line 125: We should be able to access this address by reading from the masterRegistry

src/pages/CreatePool/CreatePoolDialog.tsx
```diff
@@ -7,68 +7,172 @@ import {
...
           <Box display="flex" justifyContent="space-between">
             <Typography>{t("poolType")}</Typography>
-            <Typography>USD MetaPool</Typography>
+            <Typography>{t(poolData.poolType)}</Typography>
           </Box>
           <Box display="flex" justifyContent="space-between">
             <Typography>Tokens</Typography>
-            <Typography>vUSD</Typography>
+            <Typography>
+              {poolData.tokenInfo.map(
```
hammeiam, line 161: `poolData.tokenInfo.join(", ")`

src/pages/CreatePool/CreatePoolDialog.tsx
```diff
@@ -7,68 +7,172 @@ import {
...
-        <Typography my={3}>
-          Output is estimated. If the input is invalid or the gas is too low,
-          your transaction will revert.
-        </Typography>
+        <Typography my={3}>{outputEstimatedMsg}</Typography>
         <Stack spacing={1}>
-          <Button variant="contained" size="large" onClick={handleCreatePool}>
-            Create Pool
+          <Button variant="contained" size="large" onClick={onCreatePoolClick}>
+            <Typography>Create Pool</Typography>
```
hammeiam, line 171: i18n all these stings pls

src/pages/CreatePool/index.tsx
```diff
@@ -25,13 +26,14 @@ import ReviewCreatePool from "./CreatePoolDialog"
...
 import { getContract } from "../../utils"
 import { useActiveWeb3React } from "../../hooks"
+import { usePoolRegistry } from "../../hooks/useContract"
 import { useTranslation } from "react-i18next"
 
-type PoolType = "usdMetapool" | "btcMetapool" | "basepool"
+export type PoolType = "usdMetapool" | "btcMetapool" | "basepool"
 
-type AssetType = "USD" | "ETH" | "BTC" | "OTHERS"
+export type AssetType = 0 | 1 | 2 | 3
```
hammeiam, line 34: please use enum here. Will give you the best of both worlds `AssetType.ETH == 1`

src/pages/CreatePool/index.tsx
```diff
@@ -112,32 +137,56 @@ export default function CreatePool(): React.ReactElement {
 
   const poolNameError = poolName.length > 10
```
hammeiam, line 138: I feel like the name will be longer than the symbol almost every time - I think this limit should be the max length of bytes32

src/pages/CreatePool/index.tsx
```diff
@@ -294,12 +343,12 @@ export default function CreatePool(): React.ReactElement {
                   size="large"
                   disabled={poolType !== "basepool"}
                 >
-                  <ToggleButton value="USD" color="secondary">
+                  <ToggleButton value={0} color="secondary">
```
hammeiam, line 346: prefer enum values here eg `PoolTypes.USD`

src/pages/CreatePool/CreatePoolDialog.tsx
```diff
@@ -7,68 +7,172 @@ import {
...
+      resetFields()
+    } catch (err) {
+      console.error(err)
+      enqueueToast("error", "Unable to deploy Permissionless Pool")
+    } finally {
+      onClose()
+    }
   }
 
+  const warningMessage =
```
hammeiam, line 125: can pools be modified?

public/locales/en.json
```diff
@@ -49,7 +49,7 @@
   "addPoolNameAndSymbol": "Add pool name and symbol",
   "addPoolNameExample": "Pool name example: alETH pool, MIM Metapool.",
   "addPoolSymbolExample": "Pool Symbol example: saddle-alETH, saddle-MIM-3pool.",
-  "amplificationParameter": "Amplification parameter - controls tolerance for imbalance within the pool's assets. Higher A means lower tolerance for imbalance.",
+  "amplificationParameter": "Amplification parameter - controls tolerance for imbalance within the pool's assets. Higher A means a more stable and less change in price.",
```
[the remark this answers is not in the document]
ug02fast: Should it be `more stability and less change in price`?

package-lock.json
ug02fast: good question 😑

public/locales/en.json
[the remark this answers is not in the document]
hammeiam: Here's the tooltip text we're already using in the pool overview: `Amplification coefficient. Higher values widen the range of low-slippages swaps, while lower values help keep the pool's composition balanced.`

Maybe like `Higher A means lower slippage but is only appropriate for tightly pegged assets to avoid impermanent loss.` 🤷‍♂️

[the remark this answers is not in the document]
ug02fast: That's a good way to put it. It's also an example that is nice.

src/pages/CreatePool/CreatePoolDialog.tsx
ug02fast:
```
    updatePool(
      poolData: {
        poolAddress: string;
        lpToken: string;
        typeOfAsset: BigNumberish;
        poolName: BytesLike;
        targetAddress: string;
        tokens: string[];
        underlyingTokens: string[];
        basePoolAddress: string;
        metaSwapDepositAddress: string;
        isSaddleApproved: boolean;
        isRemoved: boolean;
        isGuarded: boolean;
      },
      overrides?: PayableOverrides & { from?: string | Promise<string> }
    ): Promise<ContractTransaction>;
```

This might do it but I'll dbl check with @penandlim

mjdatadog: This is to update pool information in the registry, I don't believe this pool data can be updated in an existing pool. Only things that we can change are a-param and fees.

ug02fast: @hammeiam The verbiage should be fine as the pool's fee and A params can be modified.

src/pages/CreatePool/index.tsx
```diff
@@ -110,34 +142,58 @@ export default function CreatePool(): React.ReactElement {
...
 
-  const poolNameError = poolName.length > 10
-  const poolSymbolError = poolSymbol.length > 14
-  const parameterError =
-    !isNumber(parameter) ||
-    parseFloat(parameter) < 1 ||
-    parseFloat(parameter) > 400
+  const poolNameError = poolName.length > 31
+  const poolSymbolError = poolSymbol.length > 31
+  const aParameterError = !isNumber(aParameter) || parseFloat(aParameter) < 1
```
hammeiam, line 147: I just looked at the `isNumber` function and there are several gaps. Most importantly, aParameter can't be a decimal, so we don't need any parseFloat logic etc. Also the `isNumber` regex itself returns true for `""`. You probably just want `/^\d+$/` and a `> 0` check

hammeiam: also `regex.test` instead of `.exec` @nkrulikovsky

src/pages/CreatePool/index.tsx
```diff
@@ -110,34 +142,58 @@ export default function CreatePool(): React.ReactElement {
...
   const feeError =
     !isNumber(fee) || parseFloat(fee) > 1 || parseFloat(fee) < 0.04
 
   useEffect(() => {
-    const ishavingErrorField =
-      poolNameError || poolSymbolError || parameterError || feeError
-    const ishavingAllValue =
+    const getBasePoolLPTokenAddrs = async () => {
+      if (poolRegistry) {
+        const basePoolName = poolType === "usdMetapool" ? "USDv2" : "BTCv2"
```
hammeiam, line 155: enums or consts

src/pages/CreatePool/index.tsx
```diff
@@ -110,34 +142,58 @@ export default function CreatePool(): React.ReactElement {
...
+          )
+          setMetapoolBasepoolAddr(poolRegistryData.poolAddress)
+          setMetapoolBasepoolLpAddr(poolRegistryData.lpToken)
+        } catch (err) {
+          console.error(err)
+        }
+      }
+    }
+    void getBasePoolLPTokenAddrs()
+  }, [poolRegistry, poolType])
```
hammeiam, line 168: this should throw a warning for missing dependency `basePoolName` and `poolRegistryData`

src/pages/CreatePool/index.tsx
```diff
@@ -110,34 +142,58 @@ export default function CreatePool(): React.ReactElement {
...
 
   useEffect(() => {
-    const ishavingErrorField =
-      poolNameError || poolSymbolError || parameterError || feeError
-    const ishavingAllValue =
+    const getBasePoolLPTokenAddrs = async () => {
+      if (poolRegistry) {
+        const basePoolName = poolType === "usdMetapool" ? "USDv2" : "BTCv2"
+        try {
+          const poolRegistryData = await poolRegistry.getPoolDataByName(
```
hammeiam, line 157: Will this fire every time the user enters a new character? If so, lets debounce

src/pages/CreatePool/index.tsx
```diff
@@ -110,34 +142,58 @@ export default function CreatePool(): React.ReactElement {
...
+            ethers.utils.formatBytes32String(basePoolName),
+          )
+          setMetapoolBasepoolAddr(poolRegistryData.poolAddress)
+          setMetapoolBasepoolLpAddr(poolRegistryData.lpToken)
+        } catch (err) {
```
hammeiam, line 162: if the registry call fails, should any state be cleared or errors raised?

src/pages/CreatePool/index.tsx
```diff
@@ -110,34 +142,58 @@ export default function CreatePool(): React.ReactElement {
...
+          console.error(err)
+        }
+      }
+    }
+    void getBasePoolLPTokenAddrs()
+  }, [poolRegistry, poolType])
+
+  useEffect(() => {
+    const tokenInfoErrors = tokenInfo.map((token) =>
+      token.checkResult === "success" ? "success" : "error",
```
hammeiam, line 172: mixing "colors" and statuses is confusing, could `TextFieldColors` just be `ValidationStates` or something instead?

src/pages/CreatePool/CreatePoolDialog.tsx
```diff
@@ -7,68 +7,155 @@ import {
...
+      decimals: BigNumberish
+      checkResult: TextFieldColors
+    }[]
+  }
+  resetFields: () => void
+  metapoolBasepoolAddr: string
+  metapoolBasepoolLpAddr: string
+}
+
+const POOL_FEE_PERCISION = 8
```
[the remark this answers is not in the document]
ug02fast: I just took that off the deployment script. I'm not sure it's correct.

src/pages/CreatePool/index.tsx
ug02fast: no, effect only runs if poolRegistry or new poolType is selected.

src/pages/CreatePool/index.tsx
ug02fast: If it fails then the state will be an empty string. Users aren't going to be able to create a permissionless pool without this and there's no user input for this value.

src/pages/CreatePool/index.tsx
ug02fast: sgtm

src/pages/CreatePool/index.tsx
```diff
@@ -322,14 +380,15 @@ export default function CreatePool(): React.ReactElement {
...
                   >
                     <TextField
+                      value={tokenInput}
                       autoComplete="off"
                       label={`Token ${index}`}
                       fullWidth
                       color={tokenInfo[index]?.checkResult ?? "primary"}
                       margin="normal"
                       onChange={(e) => {
                         tokenInputs[index] = e.target.value
```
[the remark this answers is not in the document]
ug02fast: Was having trouble keeping it uncontrolled before but seems to work now.

src/pages/CreatePool/index.tsx
ug02fast: > I just looked at the `isNumber` function and there are several gaps. Most importantly, aParameter can't be a decimal, so we don't need any parseFloat logic etc. Also the `isNumber` regex itself returns true for `""`. You probably just want `/^\d+$/` and a `> 0` check

The regex you suggested doesn't quite work right. It shows an error by default for the input field with that regex. I changed the exec to test though. The parseFloat is casting the string to a number to validate it's not less than 1. The value being passed into the contract is a BigNumber from that string.

src/pages/CreatePool/index.tsx
ug02fast: Those aren't missing dependencies here according to lint. Dependent on poolRegistry and poolType to get those values.

PR #939, opened 2022-03-31
Update review withdraw

nkrulikovsky:
## What does this PR do?
- Update ReviewWithdraw.tsx with MUI component.
- Remove `ReviewWithdraw.scss` file

## Context
https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=244%3A24898

## Screenshot

Before:

Screen Shot 2022-03-31 at 4 41 39 PM

After:

Screen Shot 2022-04-06 at 7 08 32 PM


ug02fast:
These are some things I noticed that differ from the mocks.

The modal is wider
Padding around second divider is off
Rates: has a colon
Rates are not right aligned
Total isn't displayed
Withdrawing LP token not displayed
X close button on top right isn't displayed

Can you double check Figma and Sandra on the final?

nkrulikovsky:
@ug02fast 
Thank you for letting me know these things. 
I fixed all those things but the close button will be added after modal update which I am having on another branch.

src/pages/Withdraw.tsx
```diff
@@ -35,8 +41,21 @@ function Withdraw({ poolName }: Props): ReactElement {
...
   const POOL = POOLS_MAP[poolName]
+  const [withdrawLPTokenAmount, setWithdrawLPTokenAmount] =
+    useState<BigNumber>(Zero)
+
+  const tokenInputSum = useMemo(
+    () =>
+      POOL.poolTokens.reduce(
+        (sum, { symbol }) =>
+          sum.add(
+            parseEther(withdrawFormState.tokenInputs[symbol].valueRaw) || Zero,
```
hammeiam, line 52: Does this handle parsing "" into a bignum?

src/pages/Withdraw.tsx
```diff
@@ -48,56 +67,32 @@ function Withdraw({ poolName }: Props): ReactElement {
...
-        18,
-      )
-      let withdrawLPTokenAmount
       if (poolData.totalLocked.gt(0) && tokenInputSum.gt(0)) {
+        const withdrawTokenAmounts = POOL.poolTokens.map(
+          (token) => withdrawFormState.tokenInputs[token.symbol].valueSafe,
+        )
         if (isLegacySwapABIPool(poolData.name)) {
-          withdrawLPTokenAmount = await (
+          const tokenAmount = await (
```
hammeiam, line 75: Maybe rename to "calculatedTokenAmount" to add a little more specificity

src/pages/Withdraw.tsx
nkrulikovsky: Ohhh!
It makes error.
Does it bug of ether.js? or is there any reason they do this?

nkrulikovsky: @hammeiam 

I have fixed like the following;

```
           parseEther(
              withdrawFormState.tokenInputs[symbol].valueRaw.trim() || "0",
            ) || Zero,
```

src/pages/Withdraw.tsx
nkrulikovsky: Updated!

ug02fast (approved):

sandrasong (approved):

src/pages/Withdraw.tsx
hammeiam: I think your solution of doing || "0" should work. I don't believe you'll need || Zero after it since parseEther will never evaluate to a falsey value

nkrulikovsky: @hammeiam  Before merge I didn't check your message. You are right! we can remove ||Zero.
Let me remove.

Issue #948, opened 2022-04-05
Deposit/Withdraw MyShare Card styling and layout bugs


ug02fast:
There are a few UI bugs on the MyShare card that diverges from the mockup.

Spacing below "total amount"
Bolding values
Missing Token Icons
Layout and positioning of fields.

Production:
<img width="485" alt="image" src="https://user-images.githubusercontent.com/5507707/161828745-aa868b6e-4651-4bb8-bfa5-a6e9dbe7bd63.png">

Design: 
<img width="444" alt="image" src="https://user-images.githubusercontent.com/5507707/161828791-f4db34eb-8e60-454b-972d-40af94aa27b7.png">

PR #952, opened 2022-04-06
Update PoolInforCard

nkrulikovsky:
## What does this PR do?

- update the `PoolInfoCard.tsx`.
- Remove the `PoolInfoCard.scss`.

## Context

https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=106%3A12066
close: https://github.com/saddle-finance/saddle-frontend/issues/948

## Screenshot
Screen Shot 2022-04-16 at 1 59 35 PM
Screen Shot 2022-04-16 at 1 59 47 PM
Screen Shot 2022-04-16 at 2 00 10 PM
Screen Shot 2022-04-16 at 2 00 49 PM


src/components/PoolInfoCard.tsx
```diff
@@ -61,73 +69,96 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
+
+        <InfoItem>
+          <Typography>{`${t("utilization")}:`}</Typography>
+          <Typography variant="subtitle1">
+            {formattedData.utilization}
+          </Typography>
+        </InfoItem>
+      </Box>
+      <InfoItem>
+        <Typography>{`${t("adminFee")}:`}</Typography>
```
ug02fast, line 136:
```suggestion
        <Typography>{`${t("adminFee")}: `}</Typography>
```
There needs to be a space after colon.

src/components/PoolInfoCard.tsx
```diff
@@ -61,73 +69,96 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
-          </div> */}
-        </div>
-      </div>
-      <Divider />
-      <div className="bottom">
-        <h4>{t("currencyReserves")}</h4>
-        <span>{`$${formattedData.reserve} ${t("inTotal")}`}</span>
-        <div className="tokenList">
+          <Typography variant="subtitle1">
+            :{formattedData.aParameter}
```
ug02fast, line 124: Why is it for this occurrence the colon is where the value is? There also needs to be a space. Can you move it to the above field?

src/components/PoolInfoCard.tsx
```diff
@@ -61,73 +69,96 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
+          data?.isPaused ? t("paused") : t("active")
+        }`}</Typography>
+      </InfoItem>
+      <InfoItem>
+        <Typography>{`${t("totalLocked")}:`}</Typography>
+        <Typography variant="subtitle1">{`$${formattedData.reserve}`}</Typography>
+      </InfoItem>
+      {/* <InfoItem>
+        <Typography>{t("dailyVolume") + ": "}</Typography>
+        <Typography variant="subtitle1">{formattedData.volume}</Typography>
```
ug02fast, line 101: Why is this commented out? Do we have volume data?

ug02fast (changes requested):
image
Also this needs to be LP Amount.

Please double check with Figma on the design of the smaller details.

src/components/PoolInfoCard.tsx
```diff
@@ -61,73 +69,96 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
-        <div className="infoItem">
+      <InfoItem mt={3}>
+        <Typography>{`${t("status")}:`}</Typography>
+        <Typography variant="subtitle1">{`${
+          data?.isPaused ? t("paused") : t("active")
+        }`}</Typography>
+      </InfoItem>
+      <InfoItem>
+        <Typography>{`${t("totalLocked")}:`}</Typography>
+        <Typography variant="subtitle1">{`$${formattedData.reserve}`}</Typography>
```
ug02fast, line 97:
```suggestion
        <Typography>{`${t("totalLocked")}: `} <Typography variant="subtitle1">{`${formattedData.reserve}`}</Typography></Typography>
```

Can you see if the formatting works fine with this structure? Also this had an extra `$`

ug02fast (changes requested):

src/components/PoolInfoCard.tsx
nkrulikovsky: Because tooltop is wrapping the colon as well.

src/components/PoolInfoCard.tsx
nkrulikovsky: Oh, I didn't check that there is space of 4px between property name and value

src/components/TopMenu.tsx
```diff
@@ -111,7 +111,7 @@ function TopMenu(): ReactElement {
             <NetworkDisplay onClick={handleSettingMenu} />
             <IconButton
               onClick={handleSettingMenu}
-              data-testId="settingsMenuBtn"
+              data-testid="settingsMenuBtn"
```
ug02fast, line 114: nice

ug02fast (approved):

hammeiam:
Some notes after looking at the screenshots:
- My Share USD balance should have a `$` symbol somewhere. How we represent USD should be consistent throughout the app, but it's shown in 3 different ways different between My Share, Total Locked, and Currency Reserves. 
- Something is wrong with the decimal precision of the Currency Reserves; note how the total is $173m but each asset is only in the thousands.
- The spacing of Fee and Utilization is strange, I'm not sure why they're so far to the right in the desktop layout.

hammeiam:
Could you add a style to the token icons that makes them all the same size? It looks like lpTokens are showing up a bit larger, though I don' think that was because of this PR. Also could you fix the desktop spacing of `Utilization` in the pool info panel?

Screen Shot 2022-04-13 at 12 45 55 PM
Screen Shot 2022-04-13 at 12 45 51 PM
Screen Shot 2022-04-13 at 12 45 45 PM

nkrulikovsky:
@hammeiam 
Good call out!
Gave default width and height of 24 on `TokenIcon`.
And on the design we put pool infos on two columns, but actually, virtual price is having long width than design, so it overflow.
So, my suggestion is to put pool infos on one column.

src/components/PoolInfoCard.tsx
```diff
@@ -106,37 +101,35 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
+          return formattedData.aprs[key]?.apr.gt(Zero) ? (
+            <InfoItem key={symbol}>
+              {symbol.includes("/") ? (
+                <Tooltip title={symbol.replaceAll("/", "\n")}>
+                  <Typography sx={{ borderBottom: "1px dotted" }}>
+                    Reward APR:
+                  </Typography>
+                </Tooltip>
+              ) : (
+                <Typography>{symbol} APR: &nbsp;</Typography>
```
ug02fast, line 124: Does using template literals: ` ${} ` here with a space work so you don't have to use &nbsp;?

It'd probably be `${symbol} APR: ` but with the backticks.

src/components/PoolInfoCard.tsx
```diff
@@ -181,14 +174,19 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
+                  width={20}
+                  height={20}
+                />
                 <Typography
                   variant="subtitle1"
                   ml={1}
                 >{`${token.symbol} ${token.percent}`}</Typography>
               </Box>
-              <Typography className="tokenValue">${token.value}</Typography>
+              <Typography data-testid="tokenValue">{token.value}</Typography>
```
ug02fast, line 189: Can you make this more unique to the component? This is where we will start having clashing of data-testid names. 

Maybe something like `poolInfoCardTokenValue`

src/components/WithdrawPage.tsx
```diff
@@ -225,13 +225,7 @@ const WithdrawPage = (props: Props): ReactElement => {
...
           </Button>
         </Box>
-        <Stack
-          direction="column"
-          flex={1}
-          spacing={4}
-          width="100%"
-          maxWidth="50%"
-        >
+        <Stack direction="column" flex={1} spacing={4} width="100%">
```
ug02fast, line 228: Do you always have to specify width 100%? I feel like if parent is flexing then children elements or containers should automatically fill all width?

ug02fast (changes requested):

src/components/WithdrawPage.tsx
nkrulikovsky: You are right!

src/components/PoolInfoCard.tsx
nkrulikovsky: Oh, right now this component, don't need testid actually. removed

src/components/PoolInfoCard.tsx
nkrulikovsky: Yes, it would be more pretty clear!

src/components/PoolInfoCard.tsx
```diff
@@ -28,13 +28,10 @@ const InfoItem = styled(Box)(({ theme }) => ({
   display: "flex",
   alignItems: "center",
   flexDirection: "row",
-  minWidth: "50%",
+  minWidth: "100%",
```
ug02fast, line 31: what happens is minWidth is deleted here? does the style break?

ug02fast (approved):
lgtm

src/components/PoolInfoCard.tsx
nkrulikovsky: I have added that property for putting 2 Item info on one row.
But in the case we put a Item in one row, We can remove `minWidth`

src/components/PoolInfoCard.tsx
```diff
@@ -1,24 +1,38 @@
...
-import { formatBNToPercentString, formatBNToString } from "../utils"
-import { Divider } from "@mui/material"
+import { commify, formatUnits } from "@ethersproject/units"
+import {
+  formatBNToPercentString,
+  formatBNToShortString,
+  formatBNToString,
+} from "../utils"
+
+// import { Partners } from "../utils/thirdPartyIntegrations"
```
hammeiam, line 16: rm comment

hammeiam (approved):

PR #959, opened 2022-04-07
Double Counting Balance UI Bug Fix

ug02fast:
Currently our balance summing logic already takes into account balance coming from Rewards. This adjustment to `thirdPartyIntegrations` prevents unnecessarily adding amount to balance.


hammeiam (approved):
lgtm!

PR #961, opened 2022-04-07
simplified Deposit e2e flow, asserting tx succ and value change

ug02fast:
Test was flakey. Removed wait() calls and non-essential loop logic. Now test will make a standard deposit, then assert toast renders with successful deposit message, then assert that one of the token values updates which indicates the deposit did indeed go through the flow successfully.

subtle UI format bug fix thrown in as well.


nguyenaiden (approved):
LGTM. This should reduce our flakiness. 👍

PR #964, opened 2022-04-10
Move tokenIcon references from TOKENS_MAP object to a static resource reference

nguyenaiden:
***Description:***

Currently, when a `Token` object is instantiated, the icon data is actually stored as an attribute of the object.  e.g:
`const USDC = new Token(
  USDC_CONTRACT_ADDRESSES,
  6,
  "USDC",
  "usd-coin",
  "USDC Coin",
  usdcLogo
)`
This requires us to pass the heavy icon data around to be referred in `` tags. 

***Solution:***
- Removed icon attribute from the `Token` object.
- Moved tokens icon into the public module so that they're not compiled in the final bundle
- Replaced  all `` tags across all relevant files to use the new `<TokenIcon/>` component and reference the static files along with a default grey icon fallback.
_React Component explorer view of `<TokenIcon/>`:_
image
image
resolves #964


src/components/PendingSwapExchange.tsx
```diff
@@ -131,7 +136,8 @@ const PendingSwapExchange = ({
       >
         {t("settleAsToken")}
         <img
```
hammeiam, line 138: Wow! This whole PR is beautiful. Would it be possible to create a new component like `TokenIcon` that just takes `symbol` as a prop (and arbitrary `img` element props) and handles the `getTokenIconPath` and `onError` behind the scenes?

nguyenaiden: Got you! 👍

nguyenaiden:
@nkrulikovsky, Yep! That's the goal. Lemme look into that.

hammeiam (approved):
Nicely done!

PR #966, opened 2022-04-12
Remove SwapPage.SCSS

nkrulikovsky:
## What does this PR do?
- Remove SwapPage.SCSS 
- Update swap test.

## Context
https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=221%3A22938
## Screenshot

image

image

image


src/components/SwapPage.tsx
```diff
@@ -103,7 +109,7 @@ const SwapPage = (props: Props): ReactElement => {
 
   return (
     <Container maxWidth="sm" sx={{ pt: 5, pb: 20 }}>
-      <div className="swapPage">
+      <div>
```
ug02fast, line 112: I think this can be deleted.

public/locales/en.json
```diff
@@ -102,7 +102,7 @@
   "investorClaimAmount": "Claimable: {{claimable}}",
   "investorRemainingAmount": "Total SDL: {{remaining}}",
   "language": "Language",
-  "learnMore": "Learn more",
+  "learnMore": "<Learn more>",
```
hammeiam, line 105: We would probably prefer to keep formatting out of translations

hammeiam (approved):

public/locales/en.json
nkrulikovsky: Reverted

PR #968, opened 2022-04-13
Break minichef and migrations out to their own fns

hammeiam:
The first of several steps to remove logic from usePoolData and separate concerns into their own files. Notice I'm also moving from poolname to address

### DIFF
--- a/src/utils/migrations.ts
+++ b/src/utils/migrations.ts
@@ -2,12 +2,14 @@ import {
   ChainId,
   GENERALIZED_SWAP_MIGRATOR_CONTRACT_ADDRESSES,
 } from "../constants"
-import { MulticallContract, MulticallProvider } from "../types/ethcall"
 
 import { AddressZero } from "@ethersproject/constants"
 import { Contract } from "ethcall"
 import GENERALIZED_SWAP_MIGRATOR_CONTRACT_ABI from "../constants/abis/generalizedSwapMigrator.json"
 import { GeneralizedSwapMigrator } from "../../types/ethers-contracts/GeneralizedSwapMigrator"
+import { MulticallContract } from "../types/ethcall"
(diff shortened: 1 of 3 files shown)


nkrulikovsky (approved):
Nice using of Tuple. 
LGTM!

src/utils/minichef.ts
```diff
@@ -0,0 +1,67 @@
...
+    ) as MulticallContract<MiniChef>
+
+    const [saddlePerSecond, totalAllocPoint] = await ethCallProvider.tryEach(
+      [minichefContract.saddlePerSecond(), minichefContract.totalAllocPoint()],
+      [false, false],
+    )
+
+    const poolInfos = await ethCallProvider.tryEach(
+      addressesPidTuples.map(([, pid]) => minichefContract.poolInfo(pid)),
+      Array(addressesPidTuples.length).fill(true),
```
ug02fast, line 44: I was actually trying to remember this style of filling arrays when I was refactoring multicalls to `tryEach`

ug02fast (approved):
tuple199fan

src/utils/minichef.ts
hammeiam: I feel like `calls.map(() => true)` is maybe shorter, but it all does the same thing

src/utils/migrations.ts
```diff
@@ -0,0 +1,53 @@
...
+      const result = results[i]?.newPoolAddress
+      if (result && result !== AddressZero) {
+        return { [address]: result, ...acc }
+      }
+      return acc
+    }, {} as MigrationData)
+    return parsedData
+  } catch (e) {
+    const error = new Error(`Failed to get migration data`)
+    error.stack = (e as Error).stack
```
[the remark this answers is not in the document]
hammeiam: good call out, I changed it

PR #971, opened 2022-04-13
Create providers for concern-specific data

hammeiam:
Create providers for `BasicPools`, `Tokens`, and `UserState`. BasicPools contains all on-chain data about a pool, excluding prices and user state. This currently contains much the same logic as in `usePoolData`, but will replace `usePoolData` and will itself be replaced by the Registry. Tokens contain all on-chain data about arbitrary tokens excluding user balances. UserState will contain all user state including token balances, staked amounts, claimable rewards, etc. 

In the next PR, these will be used to power `usePoolData`, `useCalculateSwapPairs`, `usePoolsStatuses` (to be killed), and the `Pools` page.


src/providers/BasicPoolsProvider.tsx
```diff
@@ -0,0 +1,208 @@
...
+} & SwapInfo
+type BasicPools = Record<string, BasicPool> | null
+
+export const BasicPoolsContext = React.createContext<BasicPools>(null)
+
+export default function BasicPoolsProvider({
+  children,
+}: React.PropsWithChildren<unknown>): ReactElement {
+  const { chainId, library } = useActiveWeb3React()
+  const [basicPools, setBasicPools] = useState<BasicPools>(null)
```
ug02fast, line 60: Is there any advantage from the UX side to providing defaults here? I'm not sure if the user will experience any improvement if there is defaults or not.

src/providers/BasicPoolsProvider.tsx
```diff
@@ -0,0 +1,208 @@
...
+export async function getSwapInfo(
+  library: Web3Provider,
+  chainId: ChainId,
+  poolName: PoolName,
+): Promise<SwapInfo | null> {
+  try {
+    // Constants
+    const pool = POOLS_MAP[poolName]
+    const _metaSwapAddress = pool.metaSwapAddresses?.[chainId]?.toLowerCase()
+    const _internalAddress = pool.addresses[chainId]?.toLowerCase() // @dev currently points to metaswapDeposit
```
ug02fast, line 125: Why are these variables underlined?

ug02fast, line 125: Why are these variables underlined?

src/providers/UserStateProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const balanceCalls = tokenAddresses.map((address) => {
+      const contract = new Contract(
+        address,
+        ERC20_ABI,
+      ) as MulticallContract<Erc20>
+      return contract.balanceOf(account)
+    })
+    const batchBalanceResults = (
+      await Promise.all(
+        batchArray(balanceCalls, BATCH_SIZE).map((batch) => {
```
ug02fast, line 84: Nice, is this the batching multicall solution? Is it working as expected?

ug02fast, line 84: Nice, is this the batching multicall solution? Is it working as expected?

ug02fast (approved):
Beautiful. Love seeing the providers.

src/providers/UserStateProvider.tsx
hammeiam: yeah it's a decent temporary solution, but basically what I had in mind

src/providers/BasicPoolsProvider.tsx
hammeiam: I was trying to denote that these variables shouldn't be used by themselves (eg directly returned like some of the other vars). I'm open to a clearer pattern for denoting this if you have one. I'm kind of borrowing from python's `_` to represent ignored variables as in `const [_, b] = getPair()`

hammeiam:
> Beautiful. Love seeing the providers. 

@ug02fast I'm still waffling a little. I actually like the "top level Updater + redux" pattern. I also think selectors can be pretty powerful.

src/providers/BasicPoolsProvider.tsx
hammeiam: This will all be replaced with registry calls, so basically if the registry calls fail, then the user can't see any pools. Thus I'm fine having the default be null instead of some half-filled pool representations

ug02fast:
> > Beautiful. Love seeing the providers.
> 
> @ug02fast I'm still waffling a little. I actually like the "top level Updater + redux" pattern. I also think selectors can be pretty powerful.

I can move all of this call logic into a react-query `useQuery` and remove the need for providers. Then use purely hooks in whatever component that smart-contract data is necessary to be in, the API will be almost identical as the data in these providers will be accessed via `useContext`. Providers feel like they've moved over to a "hold local client state that doesn't deal with async operations i.e. Modal Provider/Theme Provider/Pop Provider" type of role. This way it removes the Redux dev overhead as well as the Provider complexity and gives us server-side state management for free all leading to DX gains.

ug02fast (approved):

PR #972, opened 2022-04-13
Update TokenInput

nkrulikovsky:
## What does this PR do?
- TokenInput update
- Add error and disable props on TokenInput
- Add helptext props

## Context
https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=240%3A23923

## Screenshot
Screen Shot 2022-04-16 at 7 04 48 PM
Screen Shot 2022-04-16 at 7 05 23 PM
Screen Shot 2022-04-16 at 7 12 16 PM
Screen Shot 2022-04-16 at 7 12 36 PM


src/components/TokenInput.tsx
```diff
@@ -52,46 +63,85 @@ function TokenInput({
       periodIndex === -1 || e.target.value.length - 1 - periodIndex <= decimals
     if (isValidInput && isValidPrecision) {
       // don't allow input longer than the token allows
-      onChange(e.target.value)
+      const re = /^[0-9]*[.,]?[0-9]*$/
```
ug02fast, line 66: can you name this const to something that describes what the regex is doing? Something like `readableDecimalNumberRegex` or something.

ug02fast (changes requested):
Love the forward thinking on protecting the input from non-digit characters with the regex. Looks good, just a small suggestion on the name.

src/components/TokenInput.tsx
hammeiam: Could you also make this a const and export it from `constants/index`? That would make it easy to reuse in other components and write a couple tests for, since regex is easy to get wrong

nkrulikovsky: That would be very useful!

ug02fast (approved):

hammeiam (approved):

PR #977, opened 2022-04-21
Remove unnecessary old svgs

nkrulikovsky:
## What does this PR do?
- Remove unnecessary icons because we have replaced lots of icons with [MUI icons](https://mui.com/material-ui/material-icons/#main-content)


ug02fast (approved):
+1 if not being used

hammeiam (approved):

PR #979, opened 2022-04-22
Create minichef provider, add providers to App

hammeiam:


src/providers/TokensProvider.tsx
```diff
@@ -50,19 +52,25 @@ export default function TokensProvider({
...
+      }
       const tokenInfos = await getTokenInfos(
         ethCallProvider,
         chainId,
         Array.from(targetTokenAddresses),
       )
       if (!tokenInfos) return
       Object.keys(tokenInfos).forEach((address) => {
-        ;(tokenInfos[address] as Token).isLPToken = lpTokens.has(address)
+        ;(tokenInfos[address] as BasicToken).isLPToken = lpTokens.has(address)
```
ug02fast, line 68: what's the semicolon syntax do?

ug02fast (approved):

src/providers/TokensProvider.tsx
hammeiam: this was a lint suggestion, I think it's purpose is the same as a regular semicolon at the end of an expression, just to delimit different expressions. Seems like you'll most often see this in lines that start with `()` or `[]` since they could be used to index into a proceeding object, so the `;` makes sure their purpose is clear.

PR #980, opened 2022-04-22
Rewrite usePoolData and Pools page using new providers

hammeiam:
Makes the Pools page simple again. Rewrites usePoolData to use `basicPools`, `tokens` and `minichef` contexts, which will result in way fewer requests. 

The larger theme of this PR is we are moving away from `POOLS_MAP`, `TOKENS_MAP` and `PoolName`, which will be replaced with the aforementioned providers.


src/hooks/usePoolData.ts
```diff
@@ -101,28 +93,22 @@ const emptyPoolData = {
...
 ): PoolDataHookReturnType {
   const { account, library, chainId } = useActiveWeb3React()
-  const swapContract = useSwapContract(poolName)
-  const rewardsContract = useMiniChefContract()
-  const migratorContract = useGeneralizedSwapMigratorContract()
-  const { tokenPricesUSD, lastTransactionTimes, swapStats } = useSelector(
+  const basicPools = useContext(BasicPoolsContext)
+  const tokens = useContext(TokensContext)
+  const userState = useContext(UserStateContext)
+  const minichefData = useContext(MinichefContext)
```
ug02fast, line 108: https://github.com/dai-shi/use-context-selector#usage

might want to bring this in, in the short term as a bandaid. I'll work on a react-query based solution in the meantime.

src/hooks/usePoolData.ts
```diff
@@ -389,22 +271,68 @@ export default function usePoolData(
...
+    ),
+    18,
+  )
+  const expandedTokens = basicPool.tokens.map((token) => (tokens || {})[token])
+  const tokenBalances1e18 = basicPool.tokenBalances.map((balance, i) =>
+    balance.mul(
+      BigNumber.from(10).pow(18 - (expandedTokens[i]?.decimals || 0)),
+    ),
+  )
+  const tokenBalancesSum1e18 = tokenBalances1e18.reduce(bnSum, Zero)
```
ug02fast, line 322: hm cool

src/pages/Pools.tsx
```diff
@@ -283,58 +100,73 @@ function Pools(): ReactElement | null {
...
             }
           })
-          .map(([poolProps]) => (
+          .map((basicPool) => (
             <PoolOverview
-              key={poolProps.name}
-              {...poolProps}
+              key={basicPool.poolName}
+              poolName={basicPool.poolName}
+              poolRoute={`/pools/${basicPool.poolName}`} // TODO
```
ug02fast, line 162: should be fine, decode later if param value needs to be used somewhere.

ug02fast (commented):
looking good. like the new BN arithmetic action. maybe just use `use-context-selector` in all those providers to prevent cascading re-renders as a short term bandaid. I will work on converting those to queries.

src/pages/Pools.tsx
hammeiam: I think the risk of using raw pool names (in urls, rendering to page) is that people are sneaky and will name their pools shit like `<script>alert("🖕🖕🖕")</script>`

ug02fast: Curve seems to determine the route path off some number. I wonder if name will scale.

hammeiam: yeah they use, I believe, the pool's index in the registry for factory pools. Seems prudent. I think address or id are safe, name is trickier.

hammeiam: one other thing to mention - we'll need to keep some static routes to maintain backwards compatibility in case partners are linking to us

src/hooks/usePoolData.ts
hammeiam: This seems like redux with extra steps 🤨

nkrulikovsky (commented):
I found that this PR makes pools load fast when change route.

src/components/PoolOverview.tsx
```diff
@@ -138,7 +122,9 @@ export default function PoolOverview({
                   variant="filled"
                   size="small"
                   label={chipLabel}
-                  color={isOutdated || shouldMigrate ? "secondary" : "error"}
+                  color={
+                    poolData.isGuarded || shouldMigrate ? "secondary" : "error"
```
nguyenaiden, line 126: the `shouldMigrate` ternary is reused pretty often. Is it possible to move this into a mini private function somewhere?

src/pages/Pools.tsx
nguyenaiden: We could use a URL cleansing library or something to prevent this.

nguyenaiden (approved):
Looks good to me! 
One question, and also there are a couple of places that have copy/pasted logic
We should refactor them if you think it's worth it/possible.

src/hooks/useApproveAndMigrate.ts
```diff
@@ -28,15 +31,24 @@ export function useApproveAndMigrate(): (
   )
 
   return async function approveAndMigrate(
-    oldPoolName: PoolName | null,
+    oldPoolName: string | null,
     lpTokenBalance?: BigNumber,
   ): Promise<void> {
-    if (!migratorContract || !chainId || !account || !library || !oldPoolName)
+    const basicPool = basicPools?.[oldPoolName || ""]
```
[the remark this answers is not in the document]
hammeiam: yep, `oldPoolName` may be undefined, and TS doesn't allow you to use `undefined` to index into an object. Another way of writing this is `const basicPool = oldPoolName ? basicPools?.[oldPoolName] : null`

src/hooks/usePoolData.ts
ug02fast: It's just calling useContextSelector instead of useContext to enable context scoping to prevent re-renders. Either way there are better patterns than this and better ways to manage server-state than Redux.

src/pages/Pools.tsx
ug02fast: A dom sanitizer yeah. Would be cool to have 1:1 relationship with poolname on FE and on contract.

PR #985, opened 2022-04-25
test improvements

hammeiam:
cy.react() wait causing a 45s wait, this prevents that

Edit: 
Not saying "but it runs on my machine" but....
Screen Shot 2022-04-25 at 11 20 33 AM


ug02fast (commented):
nice find you're on a tear

cypress/integration/swap.test.ts
```diff
@@ -37,14 +41,20 @@ context("Swap Flow", () => {
         cy.get('[data-testid="dropdownContainer"]')
           .contains(poolTokenSymbols[0])
           .click()
-        cy.react("SearchSelect").should("not.exist")
+        cy.react("SearchSelect", { options: { timeout: 1000 } }).should(
```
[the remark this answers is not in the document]
hammeiam: Why wouldn't we keep it?

cypress/integration/swap.test.ts
```diff
@@ -37,14 +41,20 @@ context("Swap Flow", () => {
...
+          "not.exist",
+        )
         // cy.react("SwapTokenInput")
         //   .eq(0)
         cy.get('[data-testid="swapTokenInputFrom"]').should(
           "include.text",
           poolTokenSymbols[0],
         )
-        cy.contains("Balance:").siblings("a").should("not.have", "0.0")
```
hammeiam, line 47: this was just straight up not running before - seems like a hazard of "hot.have" commands, I think `undefined.has(foo)` will always be true

ug02fast (approved):

cypress/integration/advancedOption.test.ts
```diff
@@ -12,67 +12,65 @@ context("Advanced option test", () => {
   })
 
   function advancedOptionTest(poolName: PoolName) {
-    const host = Cypress.env("DAPP_HOST") as string
```
hammeiam, line 15: move each loop into its own test block, rm the visit and wait from each which is redundant with the beforeAll

cypress/integration/swap.test.ts
```diff
@@ -8,16 +8,21 @@ context("Swap Flow", () => {
...
       })
       it("shows all of the pool's tokens and balances in dropdown", () => {
-        cy.react("SwapTokenInput")
+        cy.get('[data-testid="swapTokenInputFrom"]')
           .eq(0)
           .contains("Choose")
           .as("dropdownButton")
           .click() // show
         poolTokenSymbols.forEach((tokenSymbol) => {
+          cy.react("ListItem", { options: { timeout: 3000 } }).should("exist") // wait for listitem to appear
```
hammeiam, line 25: assert list item presence before looking for a specific listitem

cypress/integration/swap.test.ts
```diff
@@ -94,7 +105,7 @@ context("Swap Flow", () => {
         })
       })
       it("completes a swap", () => {
-        cy.get("button").contains("Swap").click()
+        cy.get("button").contains("Swap").should("be.enabled").click()
```
hammeiam, line 108: good habit to get into for buttons that may be disabled

cypress/integration/topMenu.test.ts
```diff
@@ -1,5 +1,5 @@
 context("Top menu test", () => {
-  beforeEach(() => {
+  before(() => {
```
hammeiam, line 2: rm unnecessary beforeEach

cypress/integration/withdrawal.test.ts
```diff
@@ -34,9 +35,10 @@ context("Withdrawal Flow", () => {
   }
 
   function testPoolWithdraw(poolName: PoolName) {
-    it(`prepares the ${poolName} pool by adding assets to it`, () => {
-      basicDeposit(poolName)
-    })
+    // @dev this runs after deposit flow so is unneeded
```
hammeiam, line 38: disable redundant deposits (may need to add this back if we aggressively parallelize)

package.json
```diff
@@ -99,7 +99,7 @@
     "build-storybook": "build-storybook -s public",
     "build": "REACT_APP_GIT_SHA=`git rev-parse --short HEAD` react-scripts build",
     "bundlemon": "bundlemon --subProject no-compression --defaultCompression none && bundlemon --subProject gzip-compression --defaultCompression gzip",
-    "cy:coverage": "start-test 3000 'npx cypress-repeat run -n 5 --until-passes --rerun-failed-only'",
+    "cy:coverage": "start-test 3000 'npx cypress-repeat run -n 2 --until-passes --rerun-failed-only'",
```
hammeiam, line 102: 5 is excessive for something broken

PR #987, opened 2022-04-26
Remove poolStatuses, rewrite useCalcSwapPairs

hammeiam:
Rewrite `useCalculateSwapPairs` as part of the ongoing effort to remove static references to `POOLS_MAP` and `TOKENS_MAP`. Delete `usePoolStatuses` which is now obsolete. Refactor tests minimally.


src/utils/index.ts
```diff
@@ -24,10 +24,10 @@ import { TokenPricesUSD } from "../state/application"
 import { getAddress } from "@ethersproject/address"
 
 export function isSynthAsset(chainId: ChainId, tokenAddress: string): boolean {
-  return (
-    chainId === ChainId.MAINNET &&
-    SYNTHETIX_TOKENS.includes(tokenAddress.toLowerCase())
-  )
+  if (chainId in SYNTHETIX_TOKENS) {
+    return SYNTHETIX_TOKENS[chainId]?.includes(tokenAddress) || false
```
ug02fast, line 28: did you maybe want `return SYNTHETIX_TOKENS?.[chainId].includes(tokenAddress) || false` here?

ug02fast (approved):
App Academy Algo god work here. lgtm but not sure why you went with abstract equality checks.

src/utils/index.ts
hammeiam: Good call out! TS was being picky about this with the old types I was using, but your solution will work

src/hooks/useCalculateSwapPairs.ts
```diff
@@ -25,36 +25,73 @@ const SWAP_TYPES_ORDERED_ASC = [
...
+        const expandedUnderlyingTokens =
+          pool.underlyingTokens?.map((addr) => tokens?.[addr]) || null
+        return { ...pool, priceData, expandedTokens, expandedUnderlyingTokens }
       })
-    const tokenToPools = sortedPools.reduce((acc, { name: poolName }) => {
-      const pool = POOLS_MAP[poolName]
+      .filter((pool) => !pool.isPaused) // paused pools can't swap
+      .filter((expandedPool) => {
+        const hasExpandedTokenData = expandedPool.expandedTokens.every(
+          (token) => token != null,
```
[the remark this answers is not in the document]
hammeiam: Because it catches both null and undefined - null is usually explicitly returned, but undefined can come from `object[key]` or `object?.foo.key`. I could probably use an explicit check in more places however, I'll try and make those updates.

src/constants/index.ts
```diff
@@ -832,7 +832,7 @@ const TBTC_V2_CONTRACT_ADDRESSES = buildAddresses({
 export const TBTC_V2 = new Token(
   TBTC_V2_CONTRACT_ADDRESSES,
   18,
-  "TBTCv2",
+  "tBTC",
```
hammeiam, line 835: updating const symbols to match on-chain data

PR #988, opened 2022-04-27
veSDL - LP Staking/Unstaking in Gauge UI

ug02fast:
image

Banner and UI are going to be hidden behind a boolean flag. Will flip flag when contracts and web3 integration ready which will be done in a separate PR.


nguyenaiden (approved):
LGTM with a small comment.
Also depending on how often/how large pre-release features we will have, may be worth it to look into actual feature flag management tools like LaunchDarkly.

For the time being, thoughts on saving these feature flag variables in a single file and import them? Just so we can keep all of these conditions in the same place.

src/components/DepositPage.tsx
```diff
@@ -101,6 +105,30 @@ const DepositPage = (props: Props): ReactElement => {
...
         </Alert>
       )}
+      {false && (
+        // {myShareData?.amountsStaked.gt(Zero) && (
+        <Alert severity="error" sx={{ mb: 2 }}>
+          <Box
+            display="flex"
+            alignItems="center"
+            justifyContent="space-around"
+            sx={{ p: 0 }}
```
nkrulikovsky, line 115: Is there any reason we need to give padding 0?

src/components/DepositPage.tsx
```diff
@@ -101,6 +105,30 @@ const DepositPage = (props: Props): ReactElement => {
...
+          >
+            <Typography sx={{ p: 0 }}>
```
nkrulikovsky, line 117: I think we don't need this padding as well.

ug02fast:
> LGTM with a small comment. Also depending on how often/how large pre-release features we will have, may be worth it to look into actual feature flag management tools like LaunchDarkly.
> 
> For the time being, thoughts on saving these feature flag variables in a single file and import them? Just so we can keep all of these conditions in the same place.

LD is nice but it costs money. It also seems to be tailored more for product people who don't necessarily know how to update boolean values in the code safetly. I think as we're still small as far as team goes and responsible for product releases that it could be fine without it for now. I'm not against it but it does add some overhead.

nkrulikovsky (approved):

PR #989, opened 2022-04-27
veSDL UI 

nkrulikovsky:
WIP


src/pages/VeSDL/GaugeVote.tsx
```diff
@@ -0,0 +1,50 @@
...
+} from "@mui/material"
+import React from "react"
+import { useTranslation } from "react-i18next"
+
+export default function GaugeVote(): JSX.Element {
+  const { t } = useTranslation()
+  return (
+    <Paper sx={{ p: 2 }}>
+      <Typography variant="h2" textAlign="center">
+        Gauge Vote
```
ug02fast, line 22: all text that will be readable on UI should be i18n'd

ug02fast (changes requested):
Looking good just a few things.

ug02fast:
image
Can you make the button and the text to the left of this spread to the edge of the container?

image
These Typographies need bolding as in docs.

image
This is missing one row of info.

ug02fast:
Just a few things otherwise looks good.

src/components/TokenInput.tsx
```diff
@@ -79,7 +83,7 @@ function TokenInput({
   }
 
   return (
-    <div>
+    <div {...rest}>
```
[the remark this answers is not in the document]
nkrulikovsky: To give other props like `data-testid`, I gave this `props`.

src/components/WithdrawPage.tsx
```diff
@@ -168,7 +168,6 @@ const WithdrawPage = (props: Props): ReactElement => {
                   <TokenInput
                     key={`tokenInput-${index}`}
                     {...token}
-                    // inputValue={parseFloat(token.inputValue).toFixed(5)}
```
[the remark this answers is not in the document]
nkrulikovsky: token is having `inputValue`, and it was from `Withdraw` page, so I think we believe we don't need this

src/components/TokenInput.tsx
```diff
@@ -59,13 +63,13 @@ function TokenInput({
   }
 
   function onChangeInput(e: React.ChangeEvent<HTMLInputElement>): void {
-    const { decimals } = TOKENS_MAP[symbol]
+    const { decimals } = TOKENS_MAP[symbol] || 0
```
hammeiam, line 66: this destructing should fail, since this would evaluate to `const {decimals} = 0`. Instead do `const {decimals = 0} = TOKENS_MAP[symbol]`

src/components/TopMenu.tsx
```diff
@@ -95,6 +95,12 @@ function TopMenu(): ReactElement {
               {t("pools")}
             </NavMenu>
 
+            {false && (
```
hammeiam, line 98: You can gate this on `IS_DEVELOPMENT`

src/pages/VeSDL/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+              onChange={handleChange}
+              inputValue={sdlTokenVal}
+            />
+            <Box display="flex" alignItems="center">
+              <div>
+                <Typography mr={1} noWrap>
+                  Unlock date:
+                </Typography>
+              </div>
+              <DatePicker
```
hammeiam, line 59: lets add `minDate` equal to one day in the future.

src/pages/App.tsx
```diff
@@ -76,85 +79,91 @@ export default function App(): ReactElement {
...
-                          ))}
-                          {pools.map(({ name, route }) => (
-                            <Route
-                              exact
-                              path={`/pools/${route}/deposit`}
-                              render={(props) => (
-                                <Deposit {...props} poolName={name} />
-                              )}
-                              key={`${route}-route-deposit`}
+          <LocalizationProvider dateAdapter={AdapterDateFns}>
```
hammeiam, line 82: Would you mind moving this below the various "Data" providers? Just above `AppContainer` should be fine

src/components/TokenInput.tsx
```diff
@@ -24,25 +24,29 @@ import { useTranslation } from "react-i18next"
...
 
 function TokenInput({
   symbol,
+  name,
   max,
   inputValue,
   onChange,
   disabled,
   error,
+  allowDecimalOverflow,
```
hammeiam, line 45: what is decimal overflow in this context?

hammeiam (approved):
This is looking really good!

src/components/WithdrawPage.tsx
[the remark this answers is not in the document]
hammeiam: @ug02fast https://github.com/saddle-finance/saddle-frontend/pull/861#discussion_r813269776 :)

src/components/TokenInput.tsx
nkrulikovsky: I was going to expand the use case of `TokenInput`.
Currently, it has limited with `isValidPrecision` and this value is from `TOKEN_MAP[symbol]`, more precisely the `pool_names`.
So, to use this component for the tokens outside of pool like, SDL, and veSDL without changing much codes, I added this props.

But we can implement this logic if necessary.

src/pages/App.tsx
nkrulikovsky: Good question!
Since this `Provider` is using for `DatePicker`, so I think there will no problem if we move this above `AppContainer`!

src/pages/VeSDL/index.tsx
nkrulikovsky: Fortunately, `DatePicker `is having `minDate` by default, I think we can use this without any modify.

ug02fast:
image
Make V capitalized

ug02fast:
image
There needs to be a little bit of margin or padding above this link

nguyenaiden (approved):
1 small comment, but LGTM!

src/components/TokenInput.tsx
nkrulikovsky: I removed this variable, and instead of this variable, I added `decimalLength` props, so user can give the decimal length even though there is no token on POOL_MAP

src/components/TokenInput.tsx
```diff
@@ -59,12 +63,14 @@ function TokenInput({
...
   function onChangeInput(e: React.ChangeEvent<HTMLInputElement>): void {
-    const { decimals } = TOKENS_MAP[symbol]
+    const { decimals } = TOKENS_MAP[symbol] || { decimals: 0 }
     const parsedValue = parseFloat("0" + e.target.value)
     const periodIndex = e.target.value.indexOf(".")
     const isValidInput = e.target.value === "" || !isNaN(parsedValue)
     const isValidPrecision =
       periodIndex === -1 || e.target.value.length - 1 - periodIndex <= decimals
+    // if (isValidInput && (isValidPrecision || allowDecimalOverflow)) {
+    console.log(allowDecimalOverflow)
```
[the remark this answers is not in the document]
nkrulikovsky: Yes, I removed this variable, and added `decimalLength`.
Could you check this logic again?

src/components/TokenInput.tsx
nkrulikovsky: @ug02fast , `const { decimals } = TOKENS_MAP?.[symbol]` also gives error, when there is no token on `POOL_MAP`, because `TOKENS_MAP?.[symbol]` will be `undefined`.

ug02fast: Ok thanks for checking.

ug02fast:
> I notice when you click the top balance button it will populate the input correctly. Then when you click the bottom balance button it will Zero out the top input. The bottom balance: UI shouldn't be clickable it looks like. image

@nkrulikovsky I believe the second "veSDL" input can only display text and has no interaction. I think it's a calculation result of the first input. Can you confirm with @sandrasong ? If it's the case then make it only display the value and the balance should have no event handlers either.

ug02fast (approved):

PR #990, opened 2022-04-28
Evmos mainnet FE integration.

ug02fast:
image


hammeiam (approved):
Approving from mobile to unblock, assuming addresses are correct

PR #992, opened 2022-04-28
Support Evmos Explorer tx links.

ug02fast:
forgot to add this

### DIFF
--- a/src/components/Toastify.tsx
+++ b/src/components/Toastify.tsx
@@ -104,7 +104,9 @@ export const enqueuePromiseToast = (
               href={getMultichainScanLink(
                 chainId,
                 data?.transactionHash ?? "",
-                "tx",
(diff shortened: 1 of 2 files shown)


src/components/Toastify.tsx
```diff
@@ -104,7 +104,9 @@ export const enqueuePromiseToast = (
               href={getMultichainScanLink(
                 chainId,
                 data?.transactionHash ?? "",
-                "tx",
+                chainId === ChainId.EVMOS || chainId === ChainId.EVMOS_TESTNET
```
hammeiam, line 107: is `txs` vs `tx` an evmos idiom?

ug02fast: It is on mintscan https://www.mintscan.io/evmos/txs/11F47D7FD2AD6E43C1FFE7AB6A70D4987012B2D1F170D3B486AA92DF8EC89E44

PR #1000, opened 2022-05-03
Gauge Weight Component and Gauge Providers

nguyenaiden:
***Description***: This PR sets up the Gauge Weight PieChart along with the provider responsible for making contract calls to populate the Gauge Context with the necessary data. As of now, the Gauge data shape is:
`export type Gauge = {
  address: string
  gaugeWeight: BigNumber
  poolAddress: string
  gaugeRelativeWeight: BigNumber
  workingSupply?: BigNumber
}`
And the GaugeProvider can get `address`, `gaugeWeight`, `poolAddress`, and `gaugeRelativeWeight`. `workingSupply` to come in a later PR. 

*GaugeWeight pie chart with mock data*
image

*NOTE*: Working off of a WIP branch of the saddle-contract that has the GaugeController and the HelperContract code. Only works on HardHat.

### DIFF
--- a/src/components/GaugeWeight.tsx
+++ b/src/components/GaugeWeight.tsx
@@ -1,4 +1,5 @@
-import React, { useRef } from "react"
+import React, { useContext, useRef } from "react"
+import { BasicPoolsContext } from "../providers/BasicPoolsProvider"
 import { BigNumber } from "@ethersproject/bignumber"
 import { CircularProgress } from "@mui/material"
 import { Gauge } from "../providers/GaugeProvider"
@@ -8,7 +9,7 @@ import HighchartsReact from "highcharts-react-official"
 import PieChart from "highcharts-react-official"
 
 interface Props {
-  gauges: Gauge[] | undefined
+  gauges: { [poolAddress: string]: Gauge }
 }
 
 export default function GaugeWeight({
@@ -17,10 +18,28 @@ export default function GaugeWeight({
 }: Props & HighchartsReact.Props): JSX.Element {
   HighchartsExporting(Highcharts)
   const chartComponentRef = useRef<HighchartsReact.RefObject>(null)
-  const data = gauges?.map((g) => {
+  const pools = useContext(BasicPoolsContext)
+  if (pools == undefined) return <CircularProgress color="secondary" />
+
+  const gaugesInfo = Object.values(pools)
+    .map((pool) => {
+      if (!pool || !pool.poolAddress || !gauges[pool.poolAddress]) return
+      const gaugePoolAddress = pool.poolAddress
+      return {
+        poolAddress: gaugePoolAddress,
+        poolName: pool.poolName,
+        gaugeRelativeWeight: gauges[gaugePoolAddress].gaugeRelativeWeight,
+      }
+    })
+    .filter(
+      (gaugeInfo) =>
+        gaugeInfo != undefined && gauges[gaugeInfo.poolAddress] != null,
+    )
+
+  const data = gaugesInfo.map((g) => {
     return {
-      name: g.name,
-      y: g.gaugeRelativeWeight.div(BigNumber.from(10).pow(16)).toNumber(),
+      name: g?.poolName,
+      y: g?.gaugeRelativeWeight.div(BigNumber.from(10).pow(16)).toNumber(),
     }
   })
 
@@ -60,17 +79,13 @@ export default function GaugeWeight({
     ],
   }
 
-  if (!gauges?.length) {
-    return <CircularProgress color="secondary" />
-  } else {
-    return (
-      <PieChart
-        highcharts={Highcharts}
-        options={options}
-        ref={chartComponentRef}
-        allowChartUpdate
-        {...props}
-      />
-    )
-  }
+  return (
+    <PieChart
+      highcharts={Highcharts}
+      options={options}
+      ref={chartComponentRef}
+      allowChartUpdate
+      {...props}
+    />
+  )
 }
--- a/src/components/GaugeWeight.tsx
+++ b/src/components/GaugeWeight.tsx
@@ -1,4 +1,5 @@
 import React, { useRef } from "react"
+import { BigNumber } from "@ethersproject/bignumber"
 import { Gauge } from "../providers/GaugeProvider"
 import Highcharts from "highcharts"
 import HighchartsExporting from "highcharts/modules/exporting"
@@ -18,7 +19,7 @@ export default function GaugeWeight({
   const data = gauges?.map((g) => {
     return {
       name: g.name,
-      y: g.gaugeRelativeWeight.toNumber() / 1e16,
+      y: g.gaugeRelativeWeight.div(BigNumber.from(10).pow(16)).toNumber(),
     }
   })
 
@@ -35,7 +36,7 @@ export default function GaugeWeight({
       },
     },
     tooltip: {
-      pointFormat: "Gauge relative weights: {point.percentage:.3f}%",
+      pointFormat: "Gauge relative weights: {point.percentage:.2f}%",
     },
     plotOptions: {
       series: {
--- a/src/components/GaugeWeight.tsx
+++ b/src/components/GaugeWeight.tsx
@@ -35,22 +35,15 @@ export default function GaugeWeight({
         ? {
             poolAddress: gaugePoolAddress,
             poolName: pool.poolName,
-            // @ts-ignore: Object is possibly 'null'.
-            gaugeRelativeWeight: gauges[gaugePoolAddress].gaugeRelativeWeight,
+            gaugeRelativeWeight: gauge.gaugeRelativeWeight,
           }
         : null
     })
-    .filter(
-      (gaugeInfo) =>
-        gaugeInfo != undefined && gauges[gaugeInfo.poolAddress] != null,
-    ) as GaugeWeightData[]
+    .filter(Boolean) as GaugeWeightData[]
 
   const data = gaugesInfo.map((g) => {
     return {
-      // This check has already been done in the filter method on L34
-      // @ts-ignore: Object is possibly 'null'.
       name: g.poolName,
-      // @ts-ignore: Object is possibly 'null'.
       y: g.gaugeRelativeWeight.div(BigNumber.from(10).pow(16)).toNumber(),
     }
   })
--- a/src/components/GaugeWeight.tsx
+++ b/src/components/GaugeWeight.tsx
@@ -1,5 +1,6 @@
 import React, { useRef } from "react"
 import { BigNumber } from "@ethersproject/bignumber"
+import { CircularProgress } from "@mui/material"
 import { Gauge } from "../providers/GaugeProvider"
 import Highcharts from "highcharts"
 import HighchartsExporting from "highcharts/modules/exporting"
@@ -59,13 +60,17 @@ export default function GaugeWeight({
     ],
   }
 
-  return (
-    <PieChart
-      highcharts={Highcharts}
-      options={options}
-      ref={chartComponentRef}
-      allowChartUpdate
-      {...props}
-    />
-  )
+  if (!gauges?.length) {
+    return <CircularProgress color="secondary" />
+  } else {
+    return (
+      <PieChart
+        highcharts={Highcharts}
+        options={options}
+        ref={chartComponentRef}
+        allowChartUpdate
+        {...props}
+      />
+    )
+  }
 }
--- a/src/components/GaugeWeight.tsx
+++ b/src/components/GaugeWeight.tsx
@@ -29,8 +29,8 @@ export default function GaugeWeight({
 
   const gaugesInfo = (Object.values(pools) as BasicPool[])
     .map((pool) => {
-      const gauge = gauges[pool.poolAddress]
       const gaugePoolAddress = pool.poolAddress
+      const gauge = gauges[gaugePoolAddress]
       return gauge
         ? {
             poolAddress: gaugePoolAddress,
--- a/src/components/GaugeWeight.tsx
+++ b/src/components/GaugeWeight.tsx
@@ -1,28 +1,24 @@
 import React, { useRef } from "react"
+import { Gauge } from "../providers/GaugeProvider"
 import Highcharts from "highcharts"
 import HighchartsExporting from "highcharts/modules/exporting"
 import HighchartsReact from "highcharts-react-official"
 import PieChart from "highcharts-react-official"
 
-interface GaugeWeight {
-  name: string
-  gauge_relative_weight: string
-}
-
 interface Props {
-  gauges: GaugeWeight[]
+  gauges: Gauge[] | undefined
 }
 
-export default function Gauge({
+export default function GaugeWeight({
   gauges,
   ...props
 }: Props & HighchartsReact.Props): JSX.Element {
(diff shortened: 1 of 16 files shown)


src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const gaugeController = useGaugeController()
+  const helperContract = useHelperContract()
+  const [gauges, setGauges] = useState<Gauges>(null)
+
+  useEffect(() => {
+    async function fetchGauges() {
+      const gaugeData: Gauge[] = []
+      if (!gaugeController || !helperContract) return
+      const nGauges = await gaugeController.n_gauges()
+      for (let i = 1; i <= nGauges.toNumber(); i++) {
```
ug02fast, line 50: what is at index 0?

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,78 @@
...
+        const gaugeAddress = await gaugeController.gauges(i)
+        const gaugePoolData: GaugePoolData =
+          await helperContract.gaugeToPoolData(gaugeAddress)
+        const gaugeWeight = await gaugeController.get_gauge_weight(gaugeAddress)
```
ug02fast, line 54: Not too excited about the underscore casing. Any reason why this changed on the contract side? @penandlim @mjdatadog

ug02fast (approved):
Looking clean, good work.

src/pages/VeSDL/GaugeVote.tsx
```diff
@@ -11,17 +11,22 @@ import {
...
+  const gaugeData = useContext(GaugeContext)
   const { t } = useTranslation()
   return (
     <Paper sx={{ p: 2 }}>
       <Typography variant="h2" textAlign="center">
         {t("gaugeVote")}
       </Typography>
-      <Box height="428px"></Box>
+      <Box height="428px">
+        <GaugeWeight gauges={gaugeData?.gauges} />
```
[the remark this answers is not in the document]
nkrulikovsky: @nguyenaiden I have added the `height` because I don't have `GaugeWeight ` component.
So, I think we can remove `height`

src/providers/GaugeProvider.tsx
hammeiam: nothing, hence it gets skipped. Also we'll be moving this to a multicall once it's ready for primetime, so this loop will go away

src/providers/GaugeProvider.tsx
hammeiam: agreed. Not a deal breaker, and I understand that we're borrowing pythonic syntax along with vyper contracts

src/constants/index.ts
```diff
@@ -470,6 +470,14 @@ export const SPA_TOKEN_ADDRESSES = buildAddresses({
   [ChainId.ARBITRUM]: "0x5575552988A3A80504bBaeB1311674fCFd40aD4B",
 })
 
+export const HELPER_CONTRACT_ADDRESSES = buildAddresses({
+  [ChainId.HARDHAT]: "0x73eccD6288e117cAcA738BDAD4FEC51312166C1A",
+})
+
+export const GAUGE_CONTROLLER_ADDRESSES = buildAddresses({
```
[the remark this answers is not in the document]
hammeiam: I'm not sure if it's been added locally yet, I don't mind getting a working version, then once all the Registry stuff is fully deployed and updated in prod, doing one big const -> registry PR for addresses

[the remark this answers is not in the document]
nguyenaiden: I'll add that once @penandlim merges his changes because it's not on mainnet yet.

src/providers/BasicPoolsProvider.tsx
```diff
@@ -40,7 +40,7 @@ type SharedSwapData = {
   tokenBalances: BigNumber[]
   lpTokenSupply: BigNumber
   miniChefRewardsPid: number | null
-  isSynthetic: boolean
+  isSynthetic?: boolean
```
hammeiam, line 43: is this optional?

src/pages/VeSDL/GaugeVote.tsx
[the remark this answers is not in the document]
nguyenaiden: @nkrulikovsky, we may have to add the height so that the spinner can center correctly. Thoughts?

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const GaugeContext = React.createContext<Gauges>(initialGaugesState)
+
+export default function GaugeProvider({
+  children,
+}: React.PropsWithChildren<unknown>): ReactElement {
+  const gaugeController = useGaugeController()
+  const helperContract = useHelperContract()
+  const poolsContext = useContext(BasicPoolsContext)
```
hammeiam, line 34: Interestingly, I don't think we need to have pools a dependency of gauges. 

We can fetch `LiquidityGaugeV5.LP_TOKEN()` and use that as a key. Then whenever we need to combine pool and gauge data (eg in the pie chart, or on the pools page when displaying rewards) we can do 
```
pools.map(pool => {
  // .. do pool stuff
  if (gauges[pool.lpToken]) {
    // do gauge stuff
  }})
  ```

This is what I've been referring to when I said "the lpToken is the foreign key between pools and gauges"

File: src/pages/VeSDL/GaugeVote.tsx (hunk 3 above)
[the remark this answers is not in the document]
Comment (nkrulikovsky) [reply src/pages/VeSDL/GaugeVote.tsx:28]: @nguyenaiden 
Agreed!

File: src/providers/GaugeProvider.tsx (hunk 2 above)
Comment (ug02fast) [reply src/providers/GaugeProvider.tsx:59]: Is this Saddle's first vyper contract port?

File: src/components/GaugeWeight.tsx [hunk 7]
```diff
@@ -0,0 +1,76 @@
...
+  gauges: Gauge[] | undefined
+}
+
+export default function GaugeWeight({
+  gauges,
+  ...props
+}: Props & HighchartsReact.Props): JSX.Element {
+  HighchartsExporting(Highcharts)
+  const chartComponentRef = useRef<HighchartsReact.RefObject>(null)
+  const data = gauges?.map((g) => {
```
hammeiam, line 20:
```
const basicPools = useContext(BasicPoolContext)
const gauges = useContext(GaugesContext)

const gaugeData = Object.values(basicPools).map(pool => {
  const gauge = gauges[pool.address] // or lpToken
  if (!gauge) return null
  return {poolName: pool.poolName, weight: gauge.relative_weight} // correct this
}).filter(Boolean)
```

src/components/GaugeWeight.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return {
+      name: g.name,
+      y: g.gaugeRelativeWeight.div(BigNumber.from(10).pow(16)).toNumber(),
+    }
+  })
+
+  const options: Highcharts.Options = {
```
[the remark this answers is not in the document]
nguyenaiden: @nkrulikovsky, could I defer all the CSS/styling changes to you? This is definitely not within my scope of expertise.

src/components/GaugeWeight.tsx
hammeiam: You can also generate this data in the GaugeVote component (the parent of GaugeWeight) and just pass in the data

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+  address: string
+  gaugeWeight: BigNumber
+  poolAddress: string
+  gaugeRelativeWeight: BigNumber
+  workingSupply?: BigNumber
+}
+
+export type Gauges = {
+  gaugeCount: number
+  gauges: { [poolAddress: string]: Gauge }
```
hammeiam, line 31: pls use `Gauge | undefined`, otherwise TS will think that `gauges[anyString]` will always return a Gauge

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const initialGaugesState: Gauges = {
+  gaugeCount: 0,
```
hammeiam, line 35: I've been using `State | null` with null being the default. I'm open to having a philosophical discussion about which way is better, and adopting those changes throughout the codebase

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+}: React.PropsWithChildren<unknown>): ReactElement {
+  const { chainId, library } = useActiveWeb3React()
+  const gaugeController = useGaugeControllerContract()
+  const helperContract = useHelperContract()
+  const [gauges, setGauges] = useState<Gauges>(initialGaugesState)
+
+  useEffect(() => {
+    async function fetchGauges() {
+      if (!gaugeController || !helperContract || !chainId || !library) return
+      const nGauges = await gaugeController.n_gauges()
```
hammeiam, line 52: please wrap any network requests that might fail in a try/catch block

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+    helperContractAddress,
+    HELPER_CONTRACT_ABI,
+  )
+
+  const gaugeControllerMultiCall = createMultiCallContract<GaugeController>(
+    gaugeControllerContractAddress,
+    GAUGE_CONTROLLER_ABI,
+  )
+
+  const gaugeAddresses = await ethCallProvider.all(
```
hammeiam, line 95: please wrap any network requests that might fail in a try/catch block. See the other providers/fetcher for examples (BasicPools, Tokens, MiniChef, UserState)

src/components/GaugeWeight.tsx
```diff
@@ -0,0 +1,91 @@
...
+    })
+    .filter(
+      (gaugeInfo) =>
+        gaugeInfo != undefined && gauges[gaugeInfo.poolAddress] != null,
+    )
+
+  const data = gaugesInfo.map((g) => {
+    return {
+      name: g?.poolName,
+      y: g?.gaugeRelativeWeight.div(BigNumber.from(10).pow(16)).toNumber(),
```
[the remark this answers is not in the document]
nguyenaiden: Ack, I added the optional to avoid suppressing the ts-lint notice. But I think this is a good candidate to do so.

src/providers/GaugeProvider.tsx
hammeiam: missed this network call

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+    const gaugeRelativeWeightsPromise: Promise<BigNumber[]> =
+      ethCallProvider.all(
+        gaugeAddresses.map((gaugeAddress) =>
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call
+          gaugeControllerMultiCall.gauge_relative_weight(gaugeAddress),
+        ),
+      )
+
+    const [gaugeWeights, gaugeRelativeWeights] = await Promise.all([
```
hammeiam, line 128: nice

src/components/GaugeWeight.tsx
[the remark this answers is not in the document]
nkrulikovsky: Yep, I can implement this on another PR!

src/utils/__tests__/index.ts
```diff
@@ -210,3 +214,18 @@ describe("getTokenIconPath", () => {
...
+    const emptyAddress = "0x0000000000000000000000000000000000000000"
+    const helperContractMultiCall = createMultiCallContract<HelperContract>(
+      emptyAddress,
+      HELPER_CONTRACT_ABI,
+    )
+
+    expect(helperContractMultiCall).toBeInstanceOf(Contract)
+    expect(
+      helperContractMultiCall.gaugeToPoolAddress(emptyAddress),
+    ).toBeInstanceOf(Object)
```
hammeiam, line 229: I wonder if everything in JS is actually an instance of Object. Might be more to the point to assert that it's not a function/promise, or that it has `ethcall/Call`-specific properties

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { HelperContract } from "../../types/ethers-contracts/HelperContract"
+import { Web3Provider } from "@ethersproject/providers"
+import { useActiveWeb3React } from "../hooks"
+
+export type Gauge = {
+  address: string
+  gaugeWeight: BigNumber
+  poolAddress: string
+  gaugeRelativeWeight: BigNumber
+  workingSupply?: BigNumber
```
hammeiam, line 29: can we make this non-optional?

src/providers/GaugeProvider.tsx
```diff
@@ -0,0 +1,158 @@
...
+      helperContractAddress,
+      HELPER_CONTRACT_ABI,
+    )
+
+    const gaugeControllerMultiCall = createMultiCallContract<GaugeController>(
+      gaugeControllerContractAddress,
+      GAUGE_CONTROLLER_ABI,
+    )
+
+    const gaugeAddresses = await ethCallProvider.all(
```
hammeiam, line 100: lowercase these bad boys as well pls

src/components/GaugeWeight.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const gaugesInfo = (Object.values(pools) as BasicPool[])
+    .map((pool) => {
+      const gauge = gauges[pool.poolAddress]
+      const gaugePoolAddress = pool.poolAddress
+      return gauge
+        ? {
+            poolAddress: gaugePoolAddress,
+            poolName: pool.poolName,
+            // @ts-ignore: Object is possibly 'null'.
+            gaugeRelativeWeight: gauges[gaugePoolAddress].gaugeRelativeWeight,
```
hammeiam, line 39: `gauges[gaugePoolAddress]` is already defined as `const gauge` and has already been null-checked, pls update and remove ts-ignore

src/components/GaugeWeight.tsx
```diff
@@ -0,0 +1,103 @@
...
+          }
+        : null
+    })
+    .filter(
```
hammeiam, line 43: `filter(Boolean) as GaugeWeightData[]` please use the code I write for you 😭😭😭

src/components/GaugeWeight.tsx
```diff
@@ -0,0 +1,103 @@
...
+      (gaugeInfo) =>
+        gaugeInfo != undefined && gauges[gaugeInfo.poolAddress] != null,
+    ) as GaugeWeightData[]
+
+  const data = gaugesInfo.map((g) => {
+    return {
+      // This check has already been done in the filter method on L34
+      // @ts-ignore: Object is possibly 'null'.
```
hammeiam, line 51: this ts-ignore shouldn't be necessary

src/providers/GaugeProvider.tsx
nguyenaiden: I can, but we're going to have to initialize it with an empty value.

src/utils/__tests__/index.ts
nguyenaiden: I added positive/negative key checks. I think it should be clear.

hammeiam (approved):
might need to change gaugecontroller address later

PR #1001, opened 2022-05-05
veSDL contract integration

nkrulikovsky:
## What does this PR do?
veSDL contract integration

NOTE: 
 - `veToken calculator` will be added on another PR.
 - VotingEscrow contract didn't deployed yet, so works on hardhat only.
 - withdraw works only when SDL is enabled to transfer, so to test this repo you should add the following script which enable SDL transfer on `500_deploy_VotingEscrow.ts` file of `saddle-contract` repo
 ```
 if (await sdl.paused()) {
    try {
      await sdl.enableTransfer()
      console.log("success enable")
    } catch (error) {
      console.log("error =>", error)
    }
  }

```
## Context
https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=704%3A30779

## Screenshot

image
image
image
image

### CODE REVIEW

File: src/components/TokenInput.tsx [hunk 1]
```diff
@@ -144,6 +146,7 @@ function TokenInput({
             variant="body2"
             color={disabled ? "text.secondary" : "text.primary"}
             textAlign="end"
+            sx={{ display: showUSDprice ? "block" : "none" }}
```
ug02fast, line 149: can you wrap this component with the `showUSDprice` boolean instead of hiding it using CSS?

```
{showUSDprice && (<Component/>)}
```

src/components/TokenInput.tsx
```diff
@@ -20,6 +20,7 @@ interface Props {
   onChange?: (value: string) => void
   disabled?: boolean
   readonly?: boolean
+  showUSDprice?: boolean
```
ug02fast, line 23: if you're defaulting showUSDprice then do you still need optional flag on this type?

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
+
 export default function VeSDL(): JSX.Element {
-  const [date, setDate] = useState<Date | null>(null)
-  const [sdlTokenRawVal, setSdlTokenRawVal] = useState<string>("")
-  const [veSdlTokenRawVal] = useState<string>("")
+  const [sdlToken, setSDLToken] = useState<TokenType>({
+    maxBalance: "",
+    sdlTokenInputVal: "",
+  })
+  const [veSdlTokenVal, setVeSDLTokenVal] = useState<BigNumber>(Zero)
```
ug02fast, line 51: follow one capitalizing convention

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
+        ...prev,
+        sdlTokenInputVal: "",
+        maxBalance: formatUnits(sdlTokenBal || Zero),
+      }))
+      const vesdlBal = await votingEscrowContract?.["balanceOf(address)"](
+        account,
+      )
+      setVeSDLTokenVal(vesdlBal || Zero)
+
+      const prevLockend =
```
ug02fast, line 98: prevLockEnd

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
+        (await votingEscrowContract?.locked__end(account)) || Zero
+      setLockEnd(
+        !prevLockend.isZero() ? new Date(prevLockend.toNumber() * 1000) : null,
+      )
+
+      const vesdlToken = await votingEscrowContract?.locked(account)
```
ug02fast, line 104: veSdlToken?

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
+      setLockedSDLVal(vesdlToken?.amount || Zero)
```
ug02fast, line 105: setLockedSdlVal?

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
+    if (votingEscrowContract) {
+      const txn = await votingEscrowContract?.force_withdraw()
+      await txn.wait()
+      void fetchData()
+    }
+  }
+
+  const addLockMos = unlockDate
+    ? differenceInMonths(unlockDate, lockEnd || new Date())
+    : null
```
ug02fast, line 202: Would `0` here make more sense than `null`?

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
-      return t("increaseLockAmount", { addLockAmt: sdlTokenRawVal })
-    } else if (sdlTokenValue.gt(Zero) && addLockMos && addLockMos > 0) {
-      return t("increaseLockAmountAndTime", {
-        addLockMos,
-        addLockAmt: sdlTokenRawVal,
-      })
-    } else if (sdlTokenValue.eq(Zero) && addLockMos && addLockMos > 0) {
-      return t("increaseLockTime", { addLockMos })
+    if (lockedSDLVal.isZero()) {
+      if (sdlTokenValue.gt(Zero) && addLockMos && addLockMos > 0)
```
ug02fast, line 206: 0 is falsy so I think the addLockMos might be unnecessary here.

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,243 @@ import {
...
+      } else if (sdlTokenValue.eq(Zero) && addLockMos && addLockMos > 0) {
+        return t("increaseLockTime", { addLockMos })
+      } else {
+        return
+      }
     }
   }
 
+  const enableLock = lockedSDLVal.isZero()
+    ? sdlToken.sdlTokenInputVal === "" || unlockDate === null
```
ug02fast, line 230: this can be `!sdlToken.sdlTokenInputVal || !unlockDate` same with the below

src/pages/VeSDL/index.tsx
```diff
@@ -129,13 +330,33 @@ export default function VeSDL(): JSX.Element {
...
+            </Typography>
+            <Typography>
+              {t("totalVeSdlHolding")}: {formatUnits(veSdlTokenVal)}
+            </Typography>
+            <Alert
+              severity="error"
+              icon={false}
+              sx={{
+                textAlign: "center",
+                display: !getPenalty() ? "none" : "block",
```
ug02fast, line 348: wrap the `<Alert />` component with the `!getPenalty()` conditional instead

src/pages/VeSDL/index.tsx
```diff
@@ -68,15 +259,18 @@ export default function VeSDL(): JSX.Element {
...
                 priceUSD: 0,
               }}
-              onChange={handleChange}
-              inputValue={sdlTokenRawVal}
+              max={sdlToken.maxBalance}
+              onChange={(value) =>
+                setSDLToken((prev) => ({ ...prev, sdlTokenInputVal: value }))
+              }
+              inputValue={sdlToken.sdlTokenInputVal}
+              helperText="Exceed Token balance"
```
ug02fast, line 273: i18n

src/pages/VeSDL/index.tsx
ug02fast, line 230: same as above. the === "" and null is unnecessary if they're falsy values

ug02fast (changes requested):
looking good. just a few requests for now. not sure if I'm 100% correct on them though.

src/components/TokenInput.tsx
nkrulikovsky: Because we are using this component on withdraw and deposit page, so I defined this value as option to avoid modify other files.

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,245 @@ import {
...
+    }
+  }, [account, sdlContract, votingEscrowContract])
+
+  function calculateLockAmount() {
+    if (!unlockDate) return
+    if (isNaN(unlockDate.valueOf())) return
+    const currentTimestamp = getUnixTime(new Date())
+    const expireTimestamp = getUnixTime(unlockDate)
+    const totalAmount = sdlTokenValue.add(lockedSDLVal)
+    const roundedExpireTimestamp = BigNumber.from(expireTimestamp)
```
hammeiam, line 115: This logic rounds down - is that more appropriate than rounding up here?

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,245 @@ import {
...
+      (MULTIPLIER * leftTime) / MAXTIME,
+    ).toFixed()
+
+    const penaltyAmount = lockedSDLVal
+      .mul(BigNumber.from(penaltyRatio))
+      .div(BigNumber.from(MULTIPLIER))
+    return formatUnits(penaltyAmount)
+  }
+
+  const handleLock = async () => {
```
hammeiam, line 139: prefer to use `useCallback`

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,245 @@ import {
...
+        await txn.wait()
+        setUnlockDate(null)
+      }
+      void fetchData()
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
+  const handleUnlock = async () => {
```
hammeiam, line 193: prefer to use `useCallback`

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,245 @@ import {
...
+      .div(BigNumber.from(MULTIPLIER))
+    return formatUnits(penaltyAmount)
+  }
+
+  const handleLock = async () => {
+    if (!account || !chainId || !votingEscrowContract?.address) return
+    try {
+      const unlockTimeStamp = unlockDate ? getUnixTime(unlockDate) : null
+      if (sdlTokenValue.gt(Zero))
+        await sdlContract?.approve(votingEscrowContract?.address, sdlTokenValue)
```
hammeiam, line 144: please use `checkAndApproveTokenForTrade` to avoid asking for approval when it's not needed

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,245 @@ import {
...
+
+      if (sdlTokenValue.gt(Zero) && unlockTimeStamp) {
+        if (lockedSDLVal.isZero()) {
+          const txn = await votingEscrowContract?.create_lock(
+            sdlTokenValue,
+            unlockTimeStamp,
+          )
+          await txn.wait()
+        } else {
```
hammeiam, line 153: Do we want to allow users to both increase lock time and amount in one click? I don't want to confuse users, and I'm not sure that our current interface makes it clear that you can either add SDL or increase lock if you have an existing balance.

src/pages/VeSDL/index.tsx
hammeiam: - I think this function should be moved outside of this component and into its own function / hook. 
- It should use `enqueuePromiseToast` to notify users of the status of contract interactions
- This function should probably clear all state upon a successful contract interaction
- Furthermore, I think could be cleaned up (fewer if/elses) if you split each effect into its own function, or at least define the different conditions as constants in one place eg

```
const hasLockedSDL =  !lockedSDLVal.isZero();
const shouldCreateLock = !hasLockedSDL && sdlTokenValue.gt(Zero) && unlockTimeStamp;
const shouldIncreaseAmount = hasLockedSDL && sdlTokenValue.gt(Zero) && !unlockTimeStamp;
const shouldIncreaseAmount = hasLockedSDL && sdlTokenValue.isZero() && unlockTimeStamp;

if (shouldCreateLock) {
  enqueuePromiseToast(...)
  const txn = await create_lock(...)
  txn.wait()
  enqueuePromiseToast(...)
} else if (shouldIncreaseAmount){
} else if (shouldIncreaseLock) {
} else {
// show UI error
}

void fetchData()
resetFormState()
```

src/pages/VeSDL/index.tsx
nkrulikovsky: @penandlim said that we don't have interface to increase both token `amount` and `lock end date`. But we are having this logic on design, and @penandlim  suggested to use two transactions in this case. And I am not sure whether it will make user confused.

src/pages/VeSDL/index.tsx
nkrulikovsky: I don't think using useCallback is necessary for this function.
1. Button component is not component wrapped with useMemo, so it will re-rendered with the change of state of `VeSDL` component.
2. Also, handleUnlock is not passed as dependency in Button, so I think using useCallback is not necessary for this handler.

src/pages/VeSDL/index.tsx
nkrulikovsky: It will make code more cleaner 🔮

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,245 @@ import {
...
 import { useTranslation } from "react-i18next"
 
+type TokenType = {
+  maxBalance: string
+  sdlTokenInputVal: string
+}
+
+const WEEK = 86400 * 7
+const MAXTIME = 86400 * 365 * 4
+const MULTIPLIER = Math.pow(10, 5)
```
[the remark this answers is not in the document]
nkrulikovsky: Using Big number, I removed this value. Actually, I gave this variable for guaranteeing correctness.

public/locales/en.json
```diff
@@ -114,6 +116,7 @@
   "learnMore": "Learn more",
   "liquidityUtilization": "Liquidity utilization",
   "lockupExpiry": "Lockup Expiry",
+  "lockSdl": "Lock {{sdlAmount}} SDL for {{period}} months",
```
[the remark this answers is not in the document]
nkrulikovsky: Not sure about that. Seems like current design only want to display unit as `month`

src/pages/VeSDL/index.tsx
```diff
@@ -11,179 +11,173 @@ import {
...
-        } else {
-          const txnIncreaseAmount = await votingEscrowContract.increase_amount(
-            sdlTokenValue,
+      if (shouldCreateLock && unlockTimeStamp) {
+        const txn = await votingEscrowContract?.create_lock(
+          sdlTokenValue,
+          unlockTimeStamp,
+        )
+        await txn.wait()
+        enqueueToast("success", "Locked SDL")
```
ug02fast, line 155: can you use `enqueuePromiseToast` here? 

```
await enqueuePromiseToast(chainId, transaction.wait(), "createLock")
```

you will probably have to create a case for "createLock" to get the message correct to the user in the toast.

src/pages/VeSDL/index.tsx
```diff
@@ -11,179 +11,173 @@ import {
...
-        }
-
-        setUnlockDate(null)
-      } else if (
-        sdlTokenValue.gt(Zero) &&
-        unlockTimeStamp === null &&
-        !veSdlTokenVal.isZero()
-      ) {
+        await txnIncreaseUnlockTime.wait()
+        enqueueToast("success", "Increased unlock time")
```
ug02fast, line 166: same as prev msg.

src/pages/VeSDL/index.tsx
```diff
@@ -193,17 +187,19 @@ export default function VeSDL(): JSX.Element {
     if (votingEscrowContract) {
       const txn = await votingEscrowContract?.force_withdraw()
       await txn.wait()
+      enqueueToast("success", "Unlocked")
```
ug02fast, line 190: same as prev msg

src/pages/VeSDL/index.tsx
```diff
@@ -340,18 +336,22 @@ export default function VeSDL(): JSX.Element {
...
+                }}
+              >
+                {t("withdrawAlertMsg", {
+                  sdlValue: formatUnits(penaltyAmount),
+                })}
+              </Alert>
+            )}
             <Button
               variant="contained"
+              data-testid="unLockVeSdlBtn"
```
ug02fast, line 354: small but unlock is one word

```suggestion
              data-testid="unlockVeSdlBtn"
```

ug02fast (changes requested):

public/locales/en.json
```diff
@@ -265,6 +274,8 @@
   "btcMetapool": "BTC Metapool",
   "basepool": "Base Pool",
   "unlock": "Unlock",
+  "unlockInitiated": "Unlock initiated",
+  "unlockCompleted": "Unlock initiated",
```
ug02fast, line 278: Completed

src/pages/VeSDL/index.tsx
```diff
@@ -4,52 +4,238 @@ import {
...
+          infiniteApproval,
+        )
+      }
+
+      if (shouldCreateLock && unlockTimeStamp) {
+        const txn = await votingEscrowContract?.create_lock(
+          sdlTokenValue,
+          unlockTimeStamp,
+        )
+        void enqueuePromiseToast(chainId, txn.wait(), "createLock")
```
ug02fast, line 154: Does toast work fine with void? I didn't know that

nkrulikovsky: Because there is no anything returned from `enquePromiseToast`

ug02fast (approved):

src/pages/VeSDL/index.tsx
```diff
@@ -10,48 +10,228 @@ import {
...
+          await votingEscrowContract.increase_unlock_time(
+            BigNumber.from(unlockTimeStamp),
+          )
+        await txnIncreaseUnlockTime.wait()
+        enqueueToast("success", "Increased amount and unlock time")
+      } else if (shouldIncreaseAmount) {
+        // Deposit additional SDL into and existing lock
+        const txn = await votingEscrowContract.increase_amount(sdlTokenValue)
+        await txn.wait()
+        void enqueuePromiseToast(chainId, txn.wait(), "increaseLockAmt")
```
hammeiam, line 170: Why is `enqueueToast` used in some places, and `enqueuePromiseToast` used in others? 
Why do some blocks await `txn.wait()`, while others pass the promise directly to `enqueuePromiseToast` without await-ing?

src/pages/VeSDL/index.tsx
```diff
@@ -10,48 +10,228 @@ import {
...
-  const addLockMos = date ? differenceInMonths(date, prevUnlockDate) : null
+  const handleUnlock = async () => {
+    if (votingEscrowContract && chainId) {
+      const txn = await votingEscrowContract?.force_withdraw()
+      void enqueuePromiseToast(chainId, txn.wait(), "unlock")
+      void fetchData()
+    }
+  }
+
+  const addLockMos =
```
hammeiam, line 194: Prefer `addLockMonths`

src/pages/VeSDL/index.tsx
```diff
@@ -10,48 +10,228 @@ import {
...
+    unlockDate && !isNaN(unlockDate.valueOf())
+      ? differenceInMonths(unlockDate, lockEnd || new Date())
```
hammeiam, line 196: `unlockDate` and `lockEnd` sound like they represent the same thing, what's the difference? Is one a `newUnlockDate` or maybe a `proposedUnlockDate`?

src/pages/VeSDL/index.tsx
```diff
@@ -10,48 +10,228 @@ import {
...
-        addLockMos,
-        addLockAmt: sdlTokenRawVal,
-      })
-    } else if (sdlTokenValue.eq(Zero) && addLockMos && addLockMos > 0) {
-      return t("increaseLockTime", { addLockMos })
+    if (lockedSDLVal.isZero()) {
+      if (sdlTokenValue.gt(Zero) && addLockMos > 0)
+        return t("lockSdl", {
+          sdlAmount: sdlToken.sdlTokenInputVal,
+          period: addLockMos,
```
hammeiam, line 204: Do we need to format these months into years?

src/pages/VeSDL/index.tsx
hammeiam: Using months is fine for now, but we may change this to `weeks` before launch.

src/utils/minBigNumber.ts
```diff
@@ -0,0 +1,5 @@
+import { BigNumber } from "@ethersproject/bignumber"
+
+export function minBigNumber(a: BigNumber, b: BigNumber): BigNumber {
+  return a.lt(b) ? b : a
```
hammeiam, line 4: I think this logic is backwards. should be `min = a.lt(b) ? a : b`. Would be good to write a quick test for this

hammeiam (approved):
Approving to unblock, but I left some additional comments

src/pages/VeSDL/index.tsx
ug02fast: Enqueue promise toast captures pending and success/fail states. enqueueToast is for other scenarios when there isn't a necessarily a promise based action happening. Having both in one try/catch might be unnecessary i forget.

src/pages/VeSDL/index.tsx
nkrulikovsky: @hammeiam 
Do you want to show the date distance as the format of year, month, week? EX: 1year 2 month 2 week

hammeiam: Yes you're correct. Are we currently showing something like `Locking SDL for 200 months`?

src/pages/VeSDL/index.tsx
hammeiam: @penandlim weeks are probably the correct minimum unit for lock times, right?

Issue #811, opened 2022-01-28
Display Contract addresses for a given pool


hammeiam:
Ideally we should display as etherscan links:
Pool address
lpToken address
tokens addresses

Be aware that this will need to work across chains

https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=762%3A32570

PR #1005, opened 2022-05-10
Add pool contract address information on PoolInfoCard

nkrulikovsky:
## Description
 It's quite difficult to find the `token contracts` associated with a `pool`, especially the `LP token`.
This PR adds the token contract and LP token conract address on `PoolInfoCard` component.

## Context

https://www.figma.com/file/Aa3IfDUO68aCNr8nZTuQ8S/Product-Design-with-MUI?node-id=762%3A32576
https://saddle-finance.slack.com/archives/C019X395350/p1650985929086549

## 
closes: https://github.com/saddle-finance/saddle-frontend/issues/811

### DIFF
--- a/src/components/PoolInfoCard.tsx
+++ b/src/components/PoolInfoCard.tsx
@@ -1,4 +1,4 @@
-import { Box, Divider, Typography, styled } from "@mui/material"
+import { Box, Divider, Link, Typography, styled } from "@mui/material"
 import { POOLS_MAP, POOL_FEE_PRECISION, PoolTypes } from "../constants"
 import React, { ReactElement } from "react"
 import {
@@ -14,6 +14,7 @@ import { PoolDataType } from "../hooks/usePoolData"
 import TokenIcon from "./TokenIcon"
 import { Tooltip } from "@mui/material"
 import { Zero } from "@ethersproject/constants"
+import { shortenAddress } from "../utils/shortenAddress"
 import { useActiveWeb3React } from "../hooks"
 import { useSelector } from "react-redux"
 import { useTranslation } from "react-i18next"
@@ -40,6 +41,8 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
     POOLS_MAP?.[data.name]?.metaSwapAddresses?.[chainId] ||
     POOLS_MAP?.[data.name]?.addresses[chainId]
   )?.toLowerCase()
+  console.log("pool addr =>", poolAddress)
+  console.log("pool addr =>", data.lpToken)
   const { oneDayVolume, utilization } =
     swapStats && poolAddress in swapStats
       ? swapStats[poolAddress]
@@ -105,7 +108,9 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
           </Typography>
         </Tooltip>
       ) : (
-        <Typography variant="h1">{formattedData.name}</Typography>
+        <Typography component="span" variant="h1" mr={2}>
+          {formattedData.name}
+        </Typography>
       )}
       <InfoItem mt={3}>
         <Typography>{`${t("status")}:`}</Typography>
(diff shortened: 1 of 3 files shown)


src/components/PoolInfoCard.tsx
```diff
@@ -40,6 +41,8 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
     POOLS_MAP?.[data.name]?.metaSwapAddresses?.[chainId] ||
     POOLS_MAP?.[data.name]?.addresses[chainId]
   )?.toLowerCase()
+  console.log("pool addr =>", poolAddress)
```
hammeiam, line 44: remove

src/components/PoolInfoCard.tsx
```diff
@@ -204,6 +209,27 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
             </Box>
           ))}
         </Box>
+        <Typography mt={3}>
+          Pool address:&nbsp;
+          <Link
+            href={`https://etherscan.io/address/${poolAddress}`}
```
hammeiam, line 215: please use `getMultichainScanLink`

src/components/PoolInfoCard.tsx
```diff
@@ -204,6 +209,27 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
+            target="_blank"
+            color="inherit"
+          >
+            {shortenAddress(poolAddress)}
+          </Link>
+        </Typography>
+        <Typography mt={1}>
+          LP token address:&nbsp;
+          <Link
+            href={`https://etherscan.io/address/${data.lpToken}`}
```
hammeiam, line 225: please use `getMultichainScanLink`

hammeiam (approved):

src/components/PoolInfoCard.tsx
nkrulikovsky: Thank you.
I was going to make such function.

src/utils/getEtherscanLink.ts
```diff
@@ -5,31 +5,31 @@ export function getMultichainScanLink(
...
-      chainScanBaseName = "mintscan"
+      chainScanDomain = "evm.evmos.org"
       break
     case ChainId.KAVA_TESTNET:
-      chainScanBaseName = "evm-alpha.kava"
+      chainScanDomain = "explorer.evm-alpha.kava.io"
       break
     default:
-      chainScanBaseName = "etherscan"
+      chainScanDomain = "etherscan"
```
ug02fast, line 29: think this should be .io in this case
```suggestion
      chainScanDomain = "etherscan.io"
```

src/components/PoolInfoCard.tsx
```diff
@@ -204,6 +208,27 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
             </Box>
           ))}
         </Box>
+        <Typography mt={3}>
+          Pool address:&nbsp;
```
ug02fast, line 212: instead of using &nbps; can you just appropriate padding/margin to the component?

src/components/PoolInfoCard.tsx
```diff
@@ -204,6 +208,27 @@ function PoolInfoCard({ data }: Props): ReactElement | null {
...
+          <Link
+            href={getMultichainScanLink(chainId, poolAddress, "address")}
```
ug02fast, line 214: will all other occurrences of `getMultichainScanLink` work with this change?

ug02fast (changes requested):
looking good just a few requests

src/components/PoolInfoCard.tsx
nkrulikovsky: This function was used on `Toastify ` as well for showing transaction link.
If we use `https://evm.evmos.org/`, I believe, we can use `tx` for `type`

src/utils/getEtherscanLink.ts
```diff
@@ -5,31 +5,31 @@ export function getMultichainScanLink(
   data: string,
   type: "tx" | "token" | "address" | "block" | "txs",
 ): string {
-  let chainScanBaseName = "etherscan"
+  let chainScanDomain = "etherscan.io"
```
hammeiam, line 8: Thank you for fixing this! Could I ask you one additional favor - Could you use `src/constants/networks.ts` to power this function? That way we have all data about networks and chainScans in one place.

PR #1009, opened 2022-05-12
Stake into gauge from My Gauge Farm and Migration into gauge integrated.

ug02fast:
Contract integrations for the gauge staking functionality. Currently, there's a flag controlling rendering of the old minichef staking UI and the newer one. 

Implementation isn't polished for now until code abstractions such as hooks come to a final form.


src/components/DepositPage.tsx
```diff
@@ -49,13 +62,11 @@ interface Props {
     inputValue: string
   }>
   exceedsWallet: boolean
-  selected?: { [key: string]: any }
```
ug02fast, line 52: unused, deleted

src/components/DepositPage.tsx
```diff
@@ -1,5 +1,6 @@
 import {
   ALETH_POOL_NAME,
+  PoolName,
```
hammeiam, line 3: we can kill this, `string` is the new `PoolName`

src/components/DepositPage.tsx
```diff
@@ -71,15 +82,71 @@ const DepositPage = (props: Props): ReactElement => {
...
+    useState<LiquidityGaugeV5>()
+  const lpTokenContract = useLPTokenContract(poolData?.name as PoolName)
   const validDepositAmount = transactionData.to.totalAmount.gt(0)
   const shouldDisplayWrappedOption = isMetaPool(poolData?.name)
   const theme = useTheme()
   const isLgDown = useMediaQuery(theme.breakpoints.down("lg"))
+  const { gauges } = useContext(GaugeContext)
+
+  useEffect(() => {
+    if (!library || !account || !chainId || !poolData) return
```
hammeiam, line 100: would we want to set the state to null instead of just returning in this case?

src/components/DepositPage.tsx
```diff
@@ -71,15 +82,71 @@ const DepositPage = (props: Props): ReactElement => {
     onConfirmTransaction,
     onToggleDepositWrapped,
   } = props
-
+  const { account, chainId, library } = useActiveWeb3React()
+  const { unstake, amountStaked } = useRewardsHelpers(
```
hammeiam, line 86: Would you mind renaming these to be minichef-specific since `useRewardsHelpers` will be shared with gauges? Or is that not correct

src/components/DepositPage.tsx
```diff
@@ -71,15 +82,71 @@ const DepositPage = (props: Props): ReactElement => {
...
+      !lpTokenContract
+    )
+      return
+    try {
+      void unstake(amountStaked)
+      await checkAndApproveTokenForTrade(
+        lpTokenContract,
+        liquidityGaugeContract.address,
+        account,
+        await lpTokenContract.balanceOf(account),
```
hammeiam, line 128: I'm guessing we'd have to `await unstake` before `lpTokenContract.balanceOf(account)` increases > 0

src/components/DepositPage.tsx
```diff
@@ -71,15 +82,71 @@ const DepositPage = (props: Props): ReactElement => {
...
+        true,
```
hammeiam, line 129: why the fuck didn't I make these arguments `key: value`

src/components/MyFarm.tsx
```diff
@@ -25,18 +31,31 @@ export default function MyFarm({
     amountOfSpaClaimable,
     isPoolIncentivized,
   } = useRewardsHelpers(poolName as PoolName)
-  const { chainId } = useActiveWeb3React()
+  const lpTokenContract = useLPTokenContract(poolName as PoolName)
```
hammeiam, line 34: kill `PoolName`

src/components/MyFarm.tsx
```diff
@@ -25,18 +31,31 @@ export default function MyFarm({
...
     formatBNToString(amountStaked, 18, 4),
   )
+  const formattedLiquidityGaugeLpStakedBalance = commify(
+    formatBNToString(userGaugeBalance, 18, 4),
+  )
   const formattedSpaClaimableBalance = commify(
     formatBNToString(amountOfSpaClaimable, 18, 4),
   )
 
+  useEffect(() => {
```
hammeiam, line 50: I've got you already bruv
```
const userState = useContext(UserStateContext)
const gaugeBalance = userState.gauges[gaugeAddr].amountStaked
```

gets refreshed every 2 blocks automatically.

src/components/MyFarm.tsx
```diff
@@ -68,16 +105,40 @@ export default function MyFarm({
...
               size="large"
               fullWidth
-              disabled={amountStaked.isZero()}
-              onClick={() => unstake(amountStaked)}
+              disabled={
+                veSDLFeatureReady
+                  ? userGaugeBalance.isZero()
+                  : amountStaked.isZero()
+              }
+              onClick={
```
hammeiam, line 123: Could you move this function into a `useCallback` or just move it out of the component to make it tidier? Ditto for the `onClick` above

src/components/DepositPage.tsx
```diff
@@ -71,15 +81,72 @@ const DepositPage = (props: Props): ReactElement => {
...
+        Zero, // @dev: gas not being used
+        {
+          onTransactionError: () => {
+            throw new Error("Your transaction could not be completed")
+          },
+        },
+        chainId,
+      )
+      const txn = await liquidityGaugeContract["deposit(uint256,address,bool)"](
+        await lpTokenContract.balanceOf(account),
```
hammeiam, line 139: will this pretty much always be the same value as the last `await lpTokenContract.balanceOf(account)` in this block? can this be saved as a var to avoid calling twice?

hammeiam:
Is this feature-flagged? Is this safe to land before gauges are live?

src/components/DepositPage.tsx
ug02fast: these are diff. one is balance in the liq gauge. the deposit call is pulling balance from lpTokenContract.

ug02fast:
> Is this feature-flagged? Is this safe to land before gauges are live?

yes

src/components/DepositPage.tsx
ug02fast: One is providing a current balance of lpTokens in the user's wallet. The other is getting the user's balance of lpToken within the gauge.

PR #1013, opened 2022-05-13
Add Snapshot voting links to veSDL page

nguyenaiden:
***Description:*** As part of the veSDL Gauge component page, there will be a table with links to the snapshot voting underneath the PieChart. It will have the voting period as the title and a button that links to the snapshot voting page. The data is fetched from the Snapshot GraphQL API and added to a key in the redux store that is rendered only once on page load. 
*Notes*:
- Only 10 entries are rendered, but can be changed if need be. 
- GraphQL calls only are getting the `id, title, start, state`. Can add extra fields later if needed
- If the state of the snapshot voting is still ongoing, the Button will show up as `vote`.
- If the state of the snapshot voting is closed, the Button will show up as `results`.
- The `View all votes` links to [https://snapshot.org/#/saddlefinance.eth/](url)

Screen Shot 2022-05-13 at 1 27 42 PM

### DIFF
--- a/src/state/voteEscrowSnapshots.ts
+++ b/src/state/voteEscrowSnapshots.ts
@@ -24,15 +24,15 @@ const voteEscrowSnapshotsSlice = createSlice({
   name: "voteEscrowSnapshots",
   initialState,
   reducers: {
-    getVoteEscrowSnapshots(
+    updateVoteEscrowSnapshots(
       state,
       action: PayloadAction<VoteEscrowSnapshots>,
     ): void {
-      state.snapshots = [...state.snapshots, ...action.payload.snapshots]
+      state.snapshots = action.payload.snapshots
     },
   },
 })
 
-export const { getVoteEscrowSnapshots } = voteEscrowSnapshotsSlice.actions
+export const { updateVoteEscrowSnapshots } = voteEscrowSnapshotsSlice.actions
 
 export default voteEscrowSnapshotsSlice.reducer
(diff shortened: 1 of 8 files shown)


src/state/voteEscrowSnapshots.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const voteEscrowSnapshotsSlice = createSlice({
+  name: "voteEscrowSnapshots",
+  initialState,
+  reducers: {
+    getVoteEscrowSnapshots(
+      state,
+      action: PayloadAction<VoteEscrowSnapshots>,
+    ): void {
+      state.snapshots = [...state.snapshots, ...action.payload.snapshots]
```
hammeiam, line 31: I think you can probably just replace instead of appending

hammeiam (approved):

ug02fast (approved):

src/state/voteEscrowSnapshots.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const initialState: VoteEscrowSnapshots = {
+  snapshots: [],
+}
+
+const voteEscrowSnapshotsSlice = createSlice({
+  name: "voteEscrowSnapshots",
+  initialState,
+  reducers: {
+    getVoteEscrowSnapshots(
```
hammeiam, line 27: this reducer should be named `updateVoteEscrowSnapshots`

PR #1015, opened 2022-05-16
Claim veSDL Rewards Integration

ug02fast:


src/components/TokenClaimDialog.tsx
```diff
@@ -58,7 +66,8 @@ export default function TokenClaimDialog({
   const rewardBalances = useContext(RewardsBalancesContext)
   const {
     claimsStatuses,
-    claimPoolReward,
+    // claimPoolReward,
```
nkrulikovsky, line 69: Is this just comment?

src/components/TokenClaimDialog.tsx
```diff
@@ -128,6 +138,11 @@ export default function TokenClaimDialog({
           <img src={logo} width={138} height={138} />
         </Box>
       </Box>
+      <Box p={4}>
+        <Alert color="info" severity="warning">
+          {t("migrationGaugeWarningMsg")}
+        </Alert>
+      </Box>
```
nkrulikovsky, line 145: nit: We can give `padding ` to `Alert` directly using sx without `Box`.

```
 <Alert color="info" severity="warning" sx={{ m: 4 }}>
```

nkrulikovsky (commented):
LGTM. Just left few comments.

src/components/TokenClaimDialog.tsx
ug02fast: This will be deleted. I will revisit this depending on if we want to keep that code for any reason or not.

src/components/TokenClaimDialog.tsx
```diff
@@ -187,19 +202,41 @@ export default function TokenClaimDialog({
...
+                  {i < arr.length - 1 && <Divider key={i} />}
+                </React.Fragment>
+              ),
+          )}
+          {/* {allPoolsWithRewards.map((pool, i, arr) => ( */}
+          {/*   <React.Fragment key={pool.poolName}> */}
+          {/*     <ClaimListItem */}
+          {/*       title={pool.poolName} */}
+          {/*       amount={rewardBalances[pool.poolName] || Zero} */}
+          {/*       claimCallback={() => claimPoolReward(pool)} */}
```
ug02fast, line 232: @hammeiam safe to delete old pool list pointing to minichef?

src/components/TokenClaimDialog.tsx
```diff
@@ -49,6 +54,9 @@ export default function TokenClaimDialog({
   const { t } = useTranslation()
   const { chainId } = useActiveWeb3React()
   const basicPools = useContext(BasicPoolsContext)
+  const userState = useContext(UserStateContext)
+  const gaugeData = useContext(GaugeContext)
+  const gauges = Object.values(gaugeData.gauges)
```
hammeiam, line 59: this will be typed as `(Gauge | undefined)[]`. Since we know that anything returned by `Object.values` will be a `Gauge`, you can cast this `as Gauge[]` which should make things a little easier later on. 

The reason it's `Gauge | undefined` is because `gaugeData.gauges[someString]` may be undefined.

hammeiam: Also will this trigger rerenders? `gaugeData` will be the same (`===`) object between renders, but I'm guessing `Object.values` creates a new object each time.

src/components/TokenClaimDialog.tsx
```diff
@@ -319,6 +355,26 @@ function useRewardClaims() {
     },
     [chainId, account, rewardsContract, updateClaimStatus],
   )
+
+  const claimGaugeReward = useCallback(
+    async (gauge: Gauge) => {
+      if (!chainId || !account || !minterContract) return
+      try {
+        updateClaimStatus(gauge.poolName, STATUSES.PENDING)
+        const txn = await minterContract["mint(address)"](gauge.address)
```
hammeiam, line 364: any reason to use `["mint(address)"]` here instead of just `.mint(gauge.address)`? It doesn't look like this function is overloaded in the abi.

src/providers/TokensProvider.tsx
```diff
@@ -63,7 +62,7 @@ export default function TokensProvider({
       }
       if (gauges.gauges) {
         // add gauge tokens
-        ;(Object.values(gauges.gauges) as Gauge[]).forEach((gauge) => {
+        Object.values(gauges.gauges).forEach((gauge) => {
```
hammeiam, line 65: was that type unneeded?

src/providers/UserStateProvider.tsx
```diff
@@ -65,9 +61,7 @@ export default function UserStateProvider({
         ? getGaugeRewardsUserData(
             library,
             chainId,
-            (Object.values(gauges.gauges) as Gauge[]).map(
-              ({ address }) => address,
-            ),
+            Object.values(gauges.gauges).map(({ address }) => address),
```
hammeiam, line 64: was that type unneeded?

src/utils/gauges.ts
```diff
@@ -98,6 +100,14 @@ export async function getGaugeData(
       )
     ).map((poolAddress) => poolAddress.toLowerCase())
 
+    const gaugePoolNames: string[] = (
```
hammeiam, line 103: I kind of have mixed feelings about this... 

On one hand, having the poolName on a gauge would be convenient for some things.

On the other hand, `gaugeToPoolData` returns a bunch of stuff, is probably kind of slow, and we already have all that data in `BasicPools`. We can get all the info we need about a gauge's pool by doing something like this in the TokenClaimDialog
```
basicPools.map(pool => {
  const gauge = gaugeData.gauges[pool.poolAddress]
  if (!gauge) return null
  // do gauge stuff with full pool info
}
```

With that said, it would also be nice to be able to look up a pool by its address instead of just its name. P1.

src/components/TokenClaimDialog.tsx
```diff
@@ -319,6 +355,26 @@ function useRewardClaims() {
     },
     [chainId, account, rewardsContract, updateClaimStatus],
   )
+
+  const claimGaugeReward = useCallback(
```
hammeiam, line 359: Can this be updated to also call `claim_rewards(account)` on the gauge's `LiquidityV5` contract as well, if it has rewards? (`gaugeData.gauges[addr].rewards.length > 0`)

src/components/TokenClaimDialog.tsx
hammeiam: yep I think that should be fine! Exciting

public/locales/en.json
```diff
@@ -144,6 +144,7 @@
   "migrateToGauge": "{{farmName}} Farm upgrade! Old farm is deprecated. You are no longer getting staking rewards. Please upgrade to the gauge farm.",
   "migrationAmount": "Migration Amount",
   "migrationExplain": "This pool is outdated. We recommend migrating to the new gas optimized version to earn better returns.",
+  "migrationGaugeWarningMsg": "Farm upgrade: LP rewards have migrated to the gauge. To automatically claim your remaining rewards, go to the Deposit or Withdraw page and click the button in the top banner.",
```
hammeiam, line 147: This gets the right gist, but would you mind sharing this in a slack channel so the markops folks can help us workshop it a little? We'll probably also want to include a link to a sections of the docs that explains minichef vs Gauges in greater detail.

hammeiam (commented):
Looking really good!

src/providers/TokensProvider.tsx
hammeiam: I see this error when I try the same thing
Screen Shot 2022-05-24 at 6 55 32 PM

src/components/TokenClaimDialog.tsx
ug02fast: I was having issues with this call but it was related to having a gas property in the Minter ABI. I can change this back to `.mint`

src/providers/TokensProvider.tsx
ug02fast: I think there's some strange configuration mismatch going on with Eslint because in my workspace it gives me different results.

src/providers/UserStateProvider.tsx
ug02fast: The lint message was that typing this is redundant.

src/components/TokenClaimDialog.tsx
ug02fast: I don't see it triggering re-renders.

ug02fast: >The reason it's Gauge | undefined is because gaugeData.gauges[someString] may be undefined.

I was wondering about this.

src/utils/gauges.ts
```diff
@@ -37,7 +39,7 @@ export type GaugeReward = {
 }
 
 export type PoolAddressToGauge = {
-  [poolAddress: string]: Gauge | undefined
+  [poolAddress: string]: Gauge
```
hammeiam, line 42: no need to change this, the previous type is more acccurate because `gauges[anyAddress]` may be undefined.

Instead, you can override the type for Object.values in `react-app-env.d.ts` with

```
interface ObjectConstructor {
  keys<T>(o: T): ObjectKeys<T>
  values<T>(o: { [s: string]: T } | ArrayLike<T>): NonNullable<T>[]
  // eslint-disable-next-line @typescript-eslint/ban-types
  values(o: {}): any[]
}
```

src/utils/gauges.ts
hammeiam, line 103: revert all of this if you're going with the pools.map plan in TokenClaimDialog

hammeiam, line 103: revert all of this if you're going with the pools.map plan in TokenClaimDialog

src/components/TokenClaimDialog.tsx
```diff
@@ -49,6 +56,20 @@ export default function TokenClaimDialog({
...
+    const gauge = gaugeData.gauges[pool?.poolAddress || ""]
+    if (!gauge) return null
+    return {
+      poolName: pool?.poolName,
+      address: gauge.address,
+      rewards: gauge.rewards,
+    }
+  })
+
+  console.log({ gauges, basicPools })
```
hammeiam, line 72: rm comment

src/components/TokenClaimDialog.tsx
```diff
@@ -49,6 +56,20 @@ export default function TokenClaimDialog({
   const { t } = useTranslation()
   const { chainId } = useActiveWeb3React()
   const basicPools = useContext(BasicPoolsContext)
+  const userState = useContext(UserStateContext)
+  const gaugeData = useContext(GaugeContext)
+  // const gauges: Gauge[] = Object.values(gaugeData.gauges)
+  const gauges = Object.values(basicPools || {}).map((pool) => {
```
hammeiam, line 62: store in useMemo and pick a more specific variable name

src/components/TokenClaimDialog.tsx
```diff
@@ -72,6 +93,7 @@ export default function TokenClaimDialog({
   const formattedTotalRetroDrop = commify(
     formatBNToString(rewardBalances.retroactiveTotal, 18, 2),
   )
+
   const [allPoolsWithRewards, poolsWithUserRewards] = useMemo(() => {
```
hammeiam, line 97: adapt this to work with gauges, or delete it. We do like that it's sorted by reward balance though.

hammeiam, line 97: adapt this to work with gauges, or delete it. We do like that it's sorted by reward balance though.

hammeiam: also change `allPoolsWithRewards` on line 193 to reference your gauge data https://github.com/saddle-finance/saddle-frontend/pull/1015/files#diff-46b08b90f09d1cc35a7f88fb50620b10f86ffda551aa713fb5baa65701c95e22R193

src/components/TokenClaimDialog.tsx
```diff
@@ -187,19 +212,29 @@ export default function TokenClaimDialog({
...
-              />
-              {i < arr.length - 1 && <Divider key={i} />}
-            </React.Fragment>
-          ))}
+          {gauges.map(
+            (gauge, i, arr) =>
+              userState?.gaugeRewards?.[gauge?.address || ""]?.claimableSDL.gt(
+                Zero,
+              ) &&
+              gauge && (
```
hammeiam, line 220: shouldn't need to check for gauge existence here. You can just do a `.filter(Boolean)` when you defined `const gauges` so the array will not include any null values.

src/components/TokenClaimDialog.tsx
```diff
@@ -187,19 +212,29 @@ export default function TokenClaimDialog({
...
+                <React.Fragment key={gauge?.poolName}>
+                  <ClaimListItem
+                    title={gauge?.poolName || ""}
+                    amount={
```
hammeiam, line 224: seems easier to save this as a const since you're using it in two places within this block

src/components/TokenClaimDialog.tsx
```diff
@@ -319,6 +355,43 @@ function useRewardClaims() {
...
+      try {
+        updateClaimStatus(gauge?.poolName || "", STATUSES.PENDING)
+        if (gauge.rewards.length > 0 && gauge.address) {
+          const liquidityGaugeContract = getContract(
+            gauge.address,
+            LIQUIDITY_GAUGE_V5_ABI,
+            library,
+            account,
+          ) as LiquidityGaugeV5
+          const txn = await liquidityGaugeContract["claim_rewards(address)"](
```
hammeiam, line 375: it is possible to fire `claim_rewards` and `mint` in the same block so the user can approve 2 txns back to back instead of waiting for one to be mined before signing the other?

src/utils/gauges.ts
hammeiam: I addressed in a PR here https://github.com/saddle-finance/saddle-frontend/pull/1049

hammeiam (commented):
Logic all looks good! The only open question is how to merge this into master - hold onto it until gauges are ready, or keep the old minichef logic and keep the new stuff behind a flag. I'd vote for the latter.

src/components/TokenClaimDialog.tsx
```diff
@@ -357,31 +345,31 @@ function useRewardClaims() {
...
-          await txn.wait()
-          enqueueToast("success", "Claimed liquidity gauge rewards")
         }
-        const txn = await minterContract.mint(gauge.address)
-        await enqueuePromiseToast(chainId, txn.wait(), "claim", {
-          poolName: gauge.poolName,
-        })
+        claimPromises.push(minterContract.mint(gauge.address))
+        await Promise.all(claimPromises)
+        enqueueToast("success", "Claimed rewards")
```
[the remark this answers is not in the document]
ug02fast: big if true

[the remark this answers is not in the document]
ug02fast: works

src/components/TokenClaimDialog.tsx
```diff
@@ -94,27 +104,6 @@ export default function TokenClaimDialog({
     formatBNToString(rewardBalances.retroactiveTotal, 18, 2),
   )
 
-  const [allPoolsWithRewards, poolsWithUserRewards] = useMemo(() => {
```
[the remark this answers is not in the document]
hammeiam: I agree, cleanup needs to be part of the FF lifecycle.

src/components/TokenClaimDialog.tsx
```diff
@@ -49,6 +56,35 @@ export default function TokenClaimDialog({
...
+  const gaugesWithPoolName = useMemo<
+    ({
+      poolName: string
+      address: string
+      rewards: GaugeReward[]
+    } | null)[]
+  >(() => {
+    return Object.values(basicPools || {})
+      .map((pool) => {
+        const gauge = gaugeData.gauges[pool?.poolAddress || ""]
```
hammeiam, line 70: poolAddress should always be defined here

src/components/TokenClaimDialog.tsx
```diff
@@ -49,6 +56,35 @@ export default function TokenClaimDialog({
...
+        if (!gauge) return null
+        return {
+          poolName: pool?.poolName ?? "",
```
hammeiam, line 73: pool and poolname should always be defined here since we're mapping over pools

src/components/TokenClaimDialog.tsx
```diff
@@ -94,6 +132,8 @@ export default function TokenClaimDialog({
     return [allPoolsWithRewards, poolsWithUserRewards]
   }, [basicPools, rewardBalances])
 
+  const isVeSdlEnabled = false
```
hammeiam, line 135: use global const `IS_VESDL_LIVE`

src/components/TokenClaimDialog.tsx
```diff
@@ -188,19 +233,42 @@ export default function TokenClaimDialog({
...
-                amount={rewardBalances[pool.poolName] || Zero}
-                claimCallback={() => claimPoolReward(pool)}
-                status={
-                  claimsStatuses["allPools"] || claimsStatuses[pool.poolName]
-                }
-              />
-              {i < arr.length - 1 && <Divider key={i} />}
-            </React.Fragment>
-          ))}
+          {!isVeSdlEnabled
```
hammeiam, line 236: In light of the conversation we had on Friday about the mechanics of the token launch, this will need to be a bit more nuanced. 
The condition under which we show gauges is `IS_VESDL_LIVE` AND `chainId === ChainId.MAINNET`. We should probably create a function or hook to house this logic in case it changes (eg we also launch gauges on arbitrum)

cc @penandlim  to confirm that we're still planning to use minichef on non-mainnet to serve rewards which we voted on via minichef

src/components/TokenClaimDialog.tsx
```diff
@@ -320,6 +392,48 @@ function useRewardClaims() {
     },
     [chainId, account, rewardsContract, updateClaimStatus],
   )
+
+  const claimGaugeReward = useCallback(
+    async (
+      gauge: {
+        poolName: string | undefined
```
hammeiam, line 399: poolName should always be defined

src/components/TokenClaimDialog.tsx
ug02fast: good catch ⚾

src/components/TokenClaimDialog.tsx
penandlim: I still dont think its a good idea to have intention to support cross chain gauge right off the bat if they are not ready. Although we can use minichef on other chains, it would require multiple time sensitive coordination with our current multisig.

But if we reallllly want to do this, we can.

hammeiam (approved):

PR #1017, opened 2022-05-17
Evmos 4 Pool FE Integration

ug02fast:
[DAI, FRAX, USDC, USDT] Pool


penandlim (approved):

src/constants/index.ts
```diff
@@ -970,6 +991,7 @@ export const TBTC_POOL_TOKENS = [TBTC_V2, ...BTC_POOL_V2_TOKENS]
 export const TBTC_UNDERLYING_POOL_TOKENS = [TBTC_V2, BTC_SWAP_V2_TOKEN]
 
 export const EVMOS_BTC_POOL_TOKENS = [WBTC, RENBTC]
+export const EVMOS_4_POOL_TOKENS = [DAI, FRAX, USDC, USDT]
```
penandlim, line 994: The order of the tokens should be DAI, USDC, USDT, FRAX

https://github.com/saddle-finance/saddle-contract/pull/583/files

penandlim (changes requested):

penandlim (approved):

src/components/TokenInput.tsx
```diff
@@ -71,7 +71,7 @@ function TokenInput({
 
   return (
     <div {...rest}>
-      {max != null && (
+      {max && (
```
hammeiam, line 74: wat

ug02fast: it's safer to avoid coercion. is there a reason we need to coerce?

nkrulikovsky: I think this should be optional.

PR #1019, opened 2022-05-17
Add Gauge Rewards

hammeiam:
Adds gauge rewards to the gauge provider, as well as including them in the tokens provider for fetching token info.

To test locally, start your local evm, then run `npx hardhat run scripts/checkGaugeRewards.ts --network localhost` to create rewards 

Screen Shot 2022-05-17 at 2 42 59 PM

@ug02fast

### DIFF
--- a/src/providers/TokensProvider.tsx
+++ b/src/providers/TokensProvider.tsx
@@ -1,4 +1,5 @@
 import { BasicPool, BasicPoolsContext } from "./BasicPoolsProvider"
+import { Gauge, GaugeContext } from "./GaugeProvider"
 import {
   MulticallCall,
   MulticallContract,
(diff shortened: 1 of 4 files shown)


src/providers/TokensProvider.tsx
```diff
@@ -58,6 +60,14 @@ export default function TokensProvider({
           targetTokenAddresses.add(address)
         })
       }
+      if (gauges.gauges) {
+        // add gauge tokens
+        ;(Object.values(gauges.gauges) as Gauge[]).forEach((gauge) => {
+          gauge.rewards.forEach(({ token }) => {
```
hammeiam, line 66: minichefData pulls out all reward tokens into their own array at the top level of the object. I might want to do that here, but I don't think it makes much of a difference.

ug02fast (approved):

Issue #1018, opened 2022-05-17
[Registry] write function to fetch basicPool data from registry


hammeiam:
- should use multicalls to make as few calls as possible 
- should use registry contract
- should not use tokens_map or pools_map
- should return the same data as and be interchangeable with getPoolsBaseData defined in https://github.com/saddle-finance/saddle-frontend/blob/master/src/providers/BasicPoolsProvider.tsx#117
- basepoolsProvider should continue to use getPoolsBaseData unless a feature flag (a boolean const defined at the bottom of constants/index) is true

PR #1020, opened 2022-05-18
 Fetch basicPool data from registry

nguyenaiden:
***Description***: Currently, `BasicPoolsProvider` is using `POOLS_MAP` and `TOKENS_MAP` to get the necessary pool data. Once deprecated, the Basic Pools data will be fetched using the registry. 
`getPoolsBaseDataFromRegistry` calls the PoolRegistry contract and set the data into the `BasicPools` context.
Pools page rendered using Registry Data. Pool names are the canonical names from the Registry.
image
Resolves: https://github.com/saddle-finance/saddle-frontend/issues/1018


hammeiam:
The major pieces of feedback I have are:
- This should be using multicall way more, we should basically be multicalling each "row" of data (eg `isPaused` for all pools, `virtualPrice` for all pools). Doing `for (pool in pools)` is suboptimal.
- afik we shouldn't need to invoke any instance of SwapContract in this function (or at least we need to use it less than you are). Read though the [pool registry contract](https://github.com/saddle-finance/saddle-contract/blob/john/vesdl/contracts/registries/PoolRegistry.sol) and see that it provides getters for relevant data (eg `getVirtualPrice`, `getPaused` `getUnderlyingTokenBalances`, etc).

hammeiam:
sorry I had an old comment that I forgot to submit

src/providers/BasicPoolsProvider.tsx
```diff
@@ -111,6 +153,158 @@ export default function BasicPoolsProvider({
...
+
+  const [
+    arePoolsPaused,
+    virtualPrices,
+    swapStorages,
+    aParameters,
+    tokensBalances,
+    underlyingTokensBalances,
+    lpTokensSupplies,
+  ] = await Promise.all([
```
hammeiam, line 242: Complete data about an arbitrary number of pools in 8-ish calls, amazing :heart_eyes:

src/providers/BasicPoolsProvider.tsx
```diff
@@ -44,20 +57,23 @@ type SharedSwapData = {
...
   isSynthetic: boolean
 }
+
 type MetaSwapInfo = SharedSwapData & {
   underlyingTokens: string[]
   basePoolAddress: string
   metaSwapDepositAddress: string
   underlyingTokenBalances: BigNumber[]
-  isMetaSwap: true
+  isMetaSwap: boolean
```
hammeiam, line 66: This may mess things up downstream. Having this value be true for one type and false for another lets us do handy things like 
```
if (pool.isMetaSwap) {
  foo = pool.underlyingTokens.map(...)
}
```
without having to do `p?.underlyingTokens?.map() || []`

src/providers/BasicPoolsProvider.tsx
```diff
@@ -66,6 +82,24 @@ export type BasicPool = {
 } & SwapInfo
 export type BasicPools = { [poolName: string]: BasicPool | undefined } | null // indexed by name, which is unique in the Registry
 
+type SwapStorage = [
```
hammeiam, line 85: protip, TS also lets you do `ReturnType<typeof PoolRegistry.swapStorage>` so you can get rid of this 

https://www.typescriptlang.org/docs/handbook/utility-types.html

hammeiam (commented):
looking good so far! more comments to follow

src/providers/BasicPoolsProvider.tsx
nguyenaiden: I'll look into being able to add this back. It was causing me a ton of typing issues.

src/providers/BasicPoolsProvider.tsx
nguyenaiden: Let me look into this. Could potentially cause problems since PoolRegistry.getSwapStorage() returns a Promise.

hammeiam: Refer to ethcall.d.ts Call for how to get see the result of the Promise

hammeiam (approved):

PR #1042, opened 2022-05-23
veSDL boost calculator

nkrulikovsky:
## What does this PR do?
Calculate veSDL boost 

## Context

https://www.notion.so/saddle-finance/Gauge-Data-1919b6d6317245baa5f58cc25b17ed98#d2025bd485a641fca2730ac617dad82b

Reference : https://dao.curve.fi/minter/calc
## Screenshot

image

Screen Shot 2022-05-24 at 2 06 10 AM


src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -39,51 +70,63 @@ export default function VeTokenCalculator({
...
+          />
+          <TextField
+            label={t("depositAmount")}
+            value={userShare ? formatBNToString(userShare?.usdBalance, 18) : ""}
+          />
+          {/* TODO: should confirm the correct formula */}
+          {/* <Typography>
+            {t("maxBoost")}:{"..."}
+          </Typography> */}
           <Typography>
```
nkrulikovsky, line 109: @penandlim 
Does we have the formula to calculate the available max boost?

ug02fast:
Could you align the icon with the text closer here?
image

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -6,29 +6,60 @@ import {
...
+import { parseEther } from "@ethersproject/units"
+import usePoolData from "../../hooks/usePoolData"
 import { useTranslation } from "react-i18next"
 
-type Props = { open: boolean; onClose: () => void }
-
+type Props = {
+  open: boolean
+  onClose: () => void
+  userBalaceVeSdl: BigNumber
```
ug02fast, line 24:
```suggestion
  userBalanceVeSdl: BigNumber
```

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -6,29 +6,60 @@ import {
...
+  totalSupplyVeSdl: BigNumber
+}
+const MAXBOOST = "2.5"
```
ug02fast, line 27:
```suggestion
const MAX_BOOST = "2.5"
```

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -6,29 +6,60 @@ import {
...
 export default function VeTokenCalculator({
   open,
   onClose,
+  totalSupplyVeSdl,
+  userBalaceVeSdl,
```
ug02fast, line 32:
```suggestion
  userBalanceVeSdl,
```

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -6,29 +6,60 @@ import {
...
+    totalLPAmount,
+    userVeSdlAmount,
+    totalSupplyVeSdl,
+  )
+  const veSdlForMaxBoost =
+    totalLPAmount.gt(Zero) &&
+    userLPAmount.mul(totalSupplyVeSdl).div(totalLPAmount)
+  const boost =
+    userLPAmount.gt(Zero) &&
+    parseEther(MAXBOOST).mul(workingAmount).div(userLPAmount)
```
ug02fast, line 58:
```suggestion
    parseEther(MAX_BOOST).mul(workingAmount).div(userLPAmount)
```

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -39,51 +70,60 @@ export default function VeTokenCalculator({
...
             SelectProps={{ IconComponent: ArrowDownIcon }}
           >
             <MenuItem value="">
               <em>None</em>
             </MenuItem>
-            <MenuItem value="D4">D4 - Ethereum</MenuItem>
-            <MenuItem value="BTC">BTC - Ethereum</MenuItem>
-            <MenuItem value="tBTC">tBTC - Ethereum</MenuItem>
-            <MenuItem value="arbFrax">arbFrax - Arbitrum</MenuItem>
+            {Object.keys(POOLS_MAP).map((poolName) => (
```
ug02fast, line 90: pools hook instead of POOLS_MAP as we're currently removing all instance of POOLS_MAP

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -39,51 +70,60 @@ export default function VeTokenCalculator({
...
+            value={userShare ? formatBNToString(userShare?.usdBalance, 18) : ""}
+          />
+          <Typography>{t("maxBoost")}: 2.5</Typography>
           <Typography>
             <CalculateIcon
               color="primary"
               sx={{ verticalAlign: "text-bottom", mr: 1 }}
             />
             {t("veSdlMaxBoost")}:
+            {veSdlForMaxBoost ? formatBNToString(veSdlForMaxBoost, 18) : "..."}
```
ug02fast, line 112: There should be something other than '...' here as that's not a number.

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -39,51 +70,60 @@ export default function VeTokenCalculator({
...
           </Typography>
           <Divider />
           <Typography variant="subtitle1">{t("myBoostCalculator")}</Typography>
           <TextField
             label="My veSDL Amount"
-            value={userVeSdlAmt}
+            value={formatBNToString(
+              userVeSdlInputAmount.isZero()
+                ? userBalaceVeSdl
```
ug02fast, line 120:
```suggestion
                ? userBalanceVeSdl
```

ug02fast (changes requested):

ug02fast (approved):

src/utils/gauges.ts
```diff
@@ -116,12 +120,51 @@ export async function getGaugeData(
...
-        gaugeRelativeWeightsPromise,
-        gaugeRewardsPromise,
-      ])
+    const gaugeMulticallContracts = gaugeAddresses.map((gaugeAddress) =>
+      createMultiCallContract<LiquidityGaugeV5>(
+        gaugeAddress,
+        LIQUIDITY_GAUGE_V5_ABI,
+      ),
+    )
+    const gaugeBalancePromise = ethCallProvider.all(
```
nguyenaiden, line 129: can we use `tryAll` so that can continue if a call fails?

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -35,33 +34,60 @@ export default function VeTokenCalculator({
...
-  const totalWorkingSupply = gauge ? gauge.workingSupply : Zero
+  const boost =
+    gauge &&
+    calculateWorkingAmountAndBoost(
+      userLPAmountBN,
+      totalLPAmountBN,
+      gauge.workingBalances,
+      gauge.workingSupply,
+      parseEther(userVeSdlInputAmount || "0"),
+      parseEther(totalVeSDLInput || "0"),
```
ug02fast, line 64: one convention, either SDL or Sdl

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -99,14 +129,31 @@ export default function VeTokenCalculator({
 
           <TextField
             label={t("poolLiquidity")}
-            value={formatBNToString(totalLPAmount, 18)}
+            value={userLPAmountInput}
+            onChange={(e) => {
+              if (
+                readableDecimalNumberRegex.test(e.target.value) ||
+                e.target.value === ""
```
ug02fast, line 136: can you DRY this pattern into a constant?

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -99,14 +129,31 @@ export default function VeTokenCalculator({
...
+              ) {
+                setTotalLPAmountInput(e.target.value)
+              }
+            }}
           />
           <Typography>
-            {t("maxBoost")}: {MAX_BOOST}
+            {t("maxBoost")}:{" "}
```
ug02fast, line 155: Instead of a " " space, can you put the values into seperate Typography components and add padding/margin?

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -35,33 +34,60 @@ export default function VeTokenCalculator({
...
+  const maxBoostPossible =
+    gauge &&
+    minVeSDL &&
+    calculateWorkingAmountAndBoost(
+      userLPAmountBN,
+      totalLPAmountBN,
+      gauge.workingBalances,
+      gauge.workingSupply,
+      minVeSDL,
+      parseEther(totalVeSDLInput || "0"),
```
ug02fast, line 82: Do you worry the calculation will be off if there's a chance value can be Zero? Is it possible to null check `totalVeSDLInput` higher up or is this approach safe?

ug02fast (changes requested):
looking good just a small request

src/pages/VeSDL/VeTokenCalculator.tsx
nkrulikovsky: I think parseEther don't gives error when `totalVeSDLInput `is empty string.

ug02fast: As long as there's no scenario where it's going to default to "0" and give the wrong result for maxBoostPossible

src/components/GaugeWeight.tsx
```diff
@@ -27,7 +27,17 @@ export default function GaugeWeight({
   const chartComponentRef = useRef<HighchartsReact.RefObject>(null)
   const basicPools = useContext(BasicPoolsContext)
   const theme = useTheme()
-  if (basicPools == undefined) return <CircularProgress color="secondary" />
+  if (basicPools == undefined)
```
ug02fast, line 30: I'm sure you didn't add this but can you change this to check for all falsy values instead of explicitly checking `undefined`? 

```suggestion
  if (!basicPools)
```

ug02fast (approved):

cypress/integration/veSdl.test.ts
```diff
@@ -12,9 +12,10 @@ context("veSDL test", () => {
       .should("exist")
       .type("100")
       .should("have.value", "100")
-    cy.getBySelId("veSdlUnlockData")
-      .find("input")
-      .type("12/02/2022")
-      .should("have.value", "12/02/2022")
+
```
nguyenaiden, line 15: Was this intentional?

nguyenaiden:
A couple of minor nits. Once you've addressed @ug02fast's comments then LGTM.

cypress/integration/veSdl.test.ts
nkrulikovsky: Because this test was failed because of the iframe which is only appearing on github action, so I tested after commenting. I gave {force:true} option for type(), so it pass but I didn't find the exact reason why it is failing.

src/pages/VeSDL/VeTokenCalculator.tsx
nkrulikovsky: When `totalVeSDLInput` is null, the boost value should have 1 and  when totalVeSDLInput is Zero, it returns also 1.
So,  I think we can give Zero when it is null.

But as Duy suggested, I'll check the votingEscrowContract and gauge to avoid the `null` value.

src/pages/VeSDL/index.tsx
```diff
@@ -91,6 +92,8 @@ export default function VeSDL(): JSX.Element {
       const vesdlBal = await votingEscrowContract?.["balanceOf(address)"](
         account,
       )
+      const totalSupply = await votingEscrowContract?.["totalSupply()"]()
```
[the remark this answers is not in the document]
nkrulikovsky: OK, Good way.

I'll give skeleton when there is no votingEscrowContract.

src/pages/VeSDL/VeTokenCalculator.tsx
ug02fast: Ok thanks for explaining.

ug02fast (approved):

PR #1043, opened 2022-05-23
Try using custom healthcheck

hammeiam:


hammeiam:
worked as a healthcheck, but closed since it wasn't needed - the rest of the build steps take longer than the container needs to be ready

PR #1046, opened 2022-05-25
Add correct ethcall tryEach types based on canFail

### LINKED ISSUE #1045: Add correct nullability to ethcall type
Since the ethcall library doesn't support tuple types and just assumes all outputs are of the same type, we have defined `types/ethcall.d.ts` to add support for tuple types. 

However as multicallv3 was released, introducing `tryEach`, our internal type was updated with an incorrect type which basically just copy-pasted the existing `all` type. 

The goal of this ticket is to update the `tryEach` type to be nullable based on the value of `canFail` at the appropriate index.

hammeiam:
Before: 
```
const foo = await multicallProvider.tryEach([getName], [false]) // type: [string]
const foo = await multicallProvider.tryEach([getName], [true]) // also type: [string]
```

After: 
```
const foo = await multicallProvider.tryEach([getName], [false]) // type: [string]
const foo = await multicallProvider.tryEach([getName], [true]) //  type: [string | null]
```

Closes #1045

### DIFF
--- a/src/types/ethcall.d.ts
+++ b/src/types/ethcall.d.ts
@@ -51,38 +51,68 @@ export class MulticallProvider extends Provider {
     block?: number,
   ): Promise<[A["outputs"], B["outputs"], C["outputs"], D["outputs"]]>
   all<O>(calls: MulticallCall<I, O>[], block?: number): Promise<O[]> // fallthrough to array type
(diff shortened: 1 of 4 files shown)


nguyenaiden (approved):
+1

ug02fast (approved):

Issue #1021, opened 2022-05-18
[Registry] Fetch prices from coingecko by address rather than gecko-id


hammeiam:
We currently hardcode coingecko ids which we use to fetch their prices. As we move away from static consts for pools and tokens, we'll also need to move away from these ids, especially since we won't be able to add them for permissionless pools (in a timely manner). 

Blocker: how will this work with localhost? (it won't)

Example implementation: https://github.com/curvefi/curve-api/blob/main/utils/data/tokens-prices.js

PR #1051, opened 2022-05-25
 fetch price by address

nkrulikovsky:
## Context

https://github.com/saddle-finance/saddle-frontend/issues/1021

### DIFF
--- a/src/providers/TokensProvider.tsx
+++ b/src/providers/TokensProvider.tsx
@@ -23,7 +23,7 @@ export type BasicToken = {
   decimals: number
   isLPToken: boolean
   isSynthetic: boolean
-  typeAsset: PoolTypes
+  typeAsset?: PoolTypes
 }
 export type BasicTokens = { [address: string]: BasicToken | undefined } | null
 export const TokensContext = React.createContext<BasicTokens>(null)
@@ -44,7 +44,7 @@ export default function TokensProvider({
       }
       const ethCallProvider = await getMulticallProvider(library, chainId)
       const lpTokens = new Set()
-      let tokenType: Record<string, PoolTypes> = {}
+      const tokenType: { [tokenAddress: string]: PoolTypes | undefined } = {}
       const targetTokenAddresses = new Set(
         (Object.values(pools) as BasicPool[])
           .map((pool) => {
@@ -54,7 +54,7 @@ export default function TokensProvider({
               ...(pool.underlyingTokens || []),
               pool.lpToken,
             ]
-            tokenType = Object.assign(
+            Object.assign(
               tokenType,
               ...tokensInPool.map((address) => ({
                 [address]: pool.typeOfAsset,
--- a/src/providers/TokensProvider.tsx
+++ b/src/providers/TokensProvider.tsx
@@ -1,4 +1,5 @@
 import { BasicPool, BasicPoolsContext } from "./BasicPoolsProvider"
+import { ChainId, PoolTypes } from "../constants"
 import {
   MulticallCall,
   MulticallContract,
@@ -7,7 +8,6 @@ import {
 import React, { ReactElement, useContext, useEffect, useState } from "react"
 import { getMulticallProvider, isSynthAsset } from "../utils"
 
-import { ChainId } from "../constants"
 import { Contract } from "ethcall"
 import ERC20_ABI from "../constants/abis/erc20.json"
 import { Erc20 } from "./../../types/ethers-contracts/Erc20.d"
@@ -23,6 +23,7 @@ export type BasicToken = {
   decimals: number
   isLPToken: boolean
   isSynthetic: boolean
+  typeAsset: PoolTypes
 }
 export type BasicTokens = { [address: string]: BasicToken | undefined } | null
 export const TokensContext = React.createContext<BasicTokens>(null)
@@ -43,15 +44,23 @@ export default function TokensProvider({
       }
       const ethCallProvider = await getMulticallProvider(library, chainId)
       const lpTokens = new Set()
+      let tokenType: Record<string, PoolTypes> = {}
       const targetTokenAddresses = new Set(
         (Object.values(pools) as BasicPool[])
           .map((pool) => {
             lpTokens.add(pool.lpToken)
-            return [
+            const tokensInPool = [
               ...pool.tokens,
               ...(pool.underlyingTokens || []),
               pool.lpToken,
             ]
+            tokenType = Object.assign(
+              tokenType,
+              ...tokensInPool.map((address) => ({
+                [address]: pool.typeOfAsset,
+              })),
+            ) as Record<string, PoolTypes>
+            return tokensInPool
           })
           .flat(),
       )
@@ -77,6 +86,7 @@ export default function TokensProvider({
       if (!tokenInfos) return
       Object.keys(tokenInfos).forEach((address) => {
         ;(tokenInfos[address] as BasicToken).isLPToken = lpTokens.has(address)
+        ;(tokenInfos[address] as BasicToken).typeAsset = tokenType[address]
       })
       setTokens(tokenInfos)
     }
--- a/src/providers/TokensProvider.tsx
+++ b/src/providers/TokensProvider.tsx
@@ -1,4 +1,3 @@
-import { BasicPool, BasicPoolsContext } from "./BasicPoolsProvider"
 import { ChainId, PoolTypes } from "../constants"
 import {
   MulticallCall,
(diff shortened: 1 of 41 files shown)


src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +108,24 @@ async function getVeth2Price(
     return etherPrice
   }
 }
+
```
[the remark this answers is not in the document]
hammeiam: This is also where you would use `chainId` to detect if the network is hardhat, so you know to only fetch ETH + BTC prices.

src/providers/UserStateProvider.tsx
```diff
@@ -15,10 +15,12 @@ import { Contract } from "ethcall"
 import ERC20_ABI from "../constants/abis/erc20.json"
 import { Erc20 } from "./../../types/ethers-contracts/Erc20.d"
 import { GaugeContext } from "./GaugeProvider"
+import { IS_DEVELOPMENT } from "../utils/environment"
 import { NETWORK_NATIVE_TOKENS } from "../constants/networks"
 import { TokensContext } from "./TokensProvider"
 import { Web3Provider } from "@ethersproject/providers"
 import { Zero } from "@ethersproject/constants"
+import { getTokenPrice } from "../utils/updateTokenPrices"
```
hammeiam, line 23: I don't think this PR needs to modify the UserStateProvider at all, most of the logic should live in `updateTokenPrices`

nkrulikovsky: yeah, this was a testing to check the result!

src/utils/updateTokenPrices.ts
```diff
@@ -6,6 +6,7 @@ import { BigNumber } from "@ethersproject/bignumber"
 import SWAP_ABI from "../constants/abis/swapFlashLoan.json"
 import { SwapFlashLoan } from "../../types/ethers-contracts/SwapFlashLoan"
 import { Web3Provider } from "@ethersproject/providers"
+import { chunk } from "lodash"
```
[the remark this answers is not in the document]
nkrulikovsky: I think `lodash` is having most of useful function related with array we need to use.

src/utils/updateTokenPrices.ts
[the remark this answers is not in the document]
nkrulikovsky: This interface is really nice! I like this .

src/utils/updateTokenPrices.ts
```diff
@@ -48,7 +51,9 @@ export default function fetchTokenPricesUSD(
         tokenIds.join(","),
       )}&vs_currencies=usd
     `)
-        .then((res) => res.json())
+        .then((res) => {
```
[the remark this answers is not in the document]
nkrulikovsky: Just wanted to check the response. returned.

src/providers/TokensProvider.tsx
```diff
@@ -42,15 +43,23 @@ export default function TokensProvider({
...
         Object.values(basicPools)
           .map((pool) => {
             lpTokens.add(pool.lpToken)
-            return [
+            const tokensInPool = [
               ...pool.tokens,
               ...(pool.underlyingTokens || []),
               pool.lpToken,
             ]
+            Object.assign(
```
ug02fast, line 56: I believe you could also just do `{ tokenType, ...tokensInPool.map((address) => ({ ...etc` thought I might be wrong on this.

src/state/application.ts
```diff
@@ -16,7 +16,7 @@ interface SwapStats {
   }
 }
 export interface TokenPricesUSD {
-  [tokenSymbol: string]: number
+  [x: string]: number
```
ug02fast, line 19: What does changing name to x do? Make it more flexible?

ug02fast (approved):

src/providers/TokensProvider.tsx
hammeiam: `...foo.map` would have array keys of 0,1,2 etc, so Object.assign is the way to go

src/utils/updateTokenPrices.ts
```diff
@@ -1,17 +1,29 @@
...
 import { SwapFlashLoan } from "../../types/ethers-contracts/SwapFlashLoan"
 import { Web3Provider } from "@ethersproject/providers"
+import { chunk } from "lodash"
 import { getContract } from "./index"
 import retry from "async-retry"
-import { updateTokensPricesUSD } from "../state/application"
 
 const coinGeckoAPI = "https://api.coingecko.com/api/v3/simple/price"
 
+const nativeTokenID = ["bitcoin", "ethereum", "tether"]
```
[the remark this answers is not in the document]
hammeiam: we can just use 1 for USD

src/providers/TokensProvider.tsx
```diff
@@ -77,6 +86,7 @@ export default function TokensProvider({
       if (!tokenInfos) return
       Object.keys(tokenInfos).forEach((address) => {
         ;(tokenInfos[address] as BasicToken).isLPToken = lpTokens.has(address)
+        ;(tokenInfos[address] as BasicToken).typeAsset = tokenType[address]
```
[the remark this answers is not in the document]
hammeiam: I think it would be good to use a fake price for `OTHER` tokens on localhost, just so we can still see APRs and things like that. Can be literally any price.

I think I would prefer to default type to `OTHER` just because dealing with optional types is annoying.

src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +115,104 @@ async function getVeth2Price(
...
 }
+
+export const getTokenPrice = (
+  tokens: BasicTokens,
+  dispatch: AppDispatch,
+  chainId: ChainId,
+): void => {
+  if (!tokens) return
+  const tokenAddresses = Object.keys(tokens)
+  const platform =
```
hammeiam, line 126: I don't think we want to use `ethereum` as a default here. If we use the wrong platform for the tokens, then none of the addresses will match anyway, so we won't get any data from coingecko. This should probably log an error if platform is unknown.

src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +115,104 @@ async function getVeth2Price(
...
+    SUPPORTED_NETWORKS[chainId]?.coingeckoPlatformID || "ethereum"
+  const addressesChunk = chunk(
+    tokenAddresses,
+    MAX_ADDRESSES_PER_COINGECKO_REQUEST,
+  )
+  if (SUPPORTED_NETWORKS[chainId]) {
+    try {
+      void retry(
```
hammeiam, line 134: Doing `retry(Promise.all)` rather than `Promise.all(retry)` may result in unnecessary fetches, since you will be retrying successful fetches when one fails.

src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +115,104 @@ async function getVeth2Price(
...
+                  switch (tokens[tokenAddress]?.typeAsset) {
+                    case PoolTypes.BTC:
+                      return { [tokenAddress]: nativeTokenPrice.btc }
+                      break
+
+                    case PoolTypes.ETH:
+                      return { [tokenAddress]: nativeTokenPrice.eth }
+                      break
+                    case PoolTypes.USD:
+                      return { [tokenAddress]: nativeTokenPrice.usd }
```
hammeiam, line 197: can just use `1` here

src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +115,104 @@ async function getVeth2Price(
...
+                      break
+                    default:
```
hammeiam, line 199: for localhost I'd like us to use a placeholder value for `OTHER` or unknown tokens. Any number you want.

src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +115,104 @@ async function getVeth2Price(
...
+            }),
+        { retries: 3 },
+      )
+    } catch (error) {
+      console.error("Error on fetching price from coingecko ==>", error)
+    }
+  }
+}
+
+const arrayToHashmap = (array: (string | number)[][]) =>
```
hammeiam, line 214: Please double check if lodash has a version of this, and move it to `utils/index`

hammeiam: You can also use generics to make this type better
`(array: [K,V][]): {[key: K]: V} => ()`

src/utils/updateTokenPrices.ts
```diff
@@ -103,3 +115,104 @@ async function getVeth2Price(
     return etherPrice
   }
 }
+
+export const getTokenPrice = (
```
hammeiam, line 119: This all looks so good!!

hammeiam (approved):
Approving to unblock, just a couple small requests to change

hammeiam:
One important thing I missed in my last review - please return the prices object with `token.symbol` as they key instead of address. This maintains backwards compatibility with the rest of the codebase. We will move entirely to addresses in the future, but not now. @nkrulikovsky

src/utils/updateTokenPrices.ts
nkrulikovsky: Do you want me to give any value for `OTHER` or `unknown` token price?

src/providers/TokensProvider.tsx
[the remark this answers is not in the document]
nkrulikovsky: Oh, I got you!

PR #1055, opened 2022-05-27
Add veSDL holder fee claim

hammeiam:
I'm currently unable to test since I don't think we have any `FeeDistributor` rewards populated @penandlim


penandlim (approved):
Tested with `john/vesdl` branch on saddle-contract. LGTM

ug02fast (approved):

ug02fast (approved):

PR #1060, opened 2022-05-30
Fix Withdraw issue

nkrulikovsky:
## What does this PR do?

Preventing the white page when withdraw

## Description.
To withdraw token, when user input the percent on the withdraw page, it appears white screen.
Because LP token don't have price(it is having undefined), and it make invalide input on `commify` function. That was the reason of blank screen.


ug02fast (approved):
No more loose equality comparison please. It's notorious for causing subtle bugs like this.

PR #1064, opened 2022-06-01
Use Partial type, write getter for changeNetwork args

hammeiam:
- Use `Partial` type instead of `| undefined` for cleanliness
- Add Partial types for things that need them (TokenPrices, SwapStats)
- Add `extractAddEthereumChainArgs` which allows us to extend `SUPPORTED_NETWORKS` without running into errors


src/providers/BasicPoolsProvider.tsx
```diff
@@ -64,7 +64,7 @@ export type BasicPool = {
   isMigrated: boolean
   newPoolAddresss?: string
 } & SwapInfo
-export type BasicPools = { [poolName: string]: BasicPool | undefined } | null // indexed by name, which is unique in the Registry
+export type BasicPools = Partial<{ [poolName: string]: BasicPool }> | null // indexed by name, which is unique in the Registry
```
nkrulikovsky, line 67: I love this approach!

nkrulikovsky (approved):
LGTM!

ug02fast (approved):
Using Partials is an amazing solution to reducing having to leverage `undefined`!

PR #1071, opened 2022-06-06
Confirm unlock modal

nkrulikovsky:
## What does this PR do?

- confirm modal
- hook for confirm modal

## Context
closes: https://github.com/saddle-finance/saddle-frontend/issues/1066

## Screenshot
image


hammeiam:
> @hammeiam 
> 
> Could you confirm the text I put will be valid for "confirm unlock"? IMO, I think we need to illustrate the penalty on the confirm text, even though we displays penalty alert?

I agree the copy can be different. We want a warning, something like "are you sure you want to unlock early? You will only receive x% of your locked sdl and you will lose the rest"

src/state/user.ts
```diff
@@ -28,10 +28,19 @@ export enum Deadlines {
   Custom = "CUSTOM",
 }
 
+export type ConfirmModalType = {
+  open: boolean
+  options?: {
+    modalTitle?: string
+    modalText?: string
+    onOK?: () => void
+    onCancel?: () => void
```
hammeiam, line 37: I don't know if you can safely persist this to user state. The user state is written to local storage and is first serialized to json. The app state then deserializes from local storage to hydrate, so the function wouldn't work

src/state/user.ts
```diff
@@ -28,10 +28,19 @@ export enum Deadlines {
   Custom = "CUSTOM",
 }
 
+export type ConfirmModalType = {
+  open: boolean
+  options?: {
+    modalTitle?: string
+    modalText?: string
+    onOK?: () => void
```
hammeiam, line 36: Are we doing anything to prevent this state from being stale? It seems like it's possible for Flow X to accidentally call onOK that was enqueued by Flow Y, which would be problematic. 

I think these should either be namespaced, or somehow cleared between flows (although multiple flows may exist on one page, which is another problem).

I think what you will end up with is some version of a Modal Controller with an interface similar to https://www.npmjs.com/package/react-modal-controller

hammeiam (approved):

PR #1072, opened 2022-06-08
Took a stab at improvements

hammeiam:
- Shorter pools multicalls impl. 
- Fix var name. 
- Improve types. 
- Sort stuff by name.

### DIFF
--- a/src/providers/BasicPoolsProvider.tsx
+++ b/src/providers/BasicPoolsProvider.tsx
@@ -8,7 +8,12 @@ import {
   getMinichefPid,
 } from "../constants"
 import {
-  DeepNullable,
+  MulticallCall,
+  MulticallContract,
+  MulticallProvider,
+} from "../types/ethcall"
+import React, { ReactElement, useEffect, useState } from "react"
+import {
   chunkedTryAll,
   createMultiCallContract,
   enumerate,
@@ -17,14 +22,8 @@ import {
   isSynthAsset,
   mapToLowerCase,
 } from "../utils"
-import {
-  MulticallCall,
-  MulticallContract,
-  MulticallProvider,
-} from "../types/ethcall"
-
-import React, { ReactElement, useEffect, useState } from "react"
 import { usePoolRegistry, usePoolRegistryMultiCall } from "../hooks/useContract"
+
 import { AppState } from "../state"
 import { BigNumber } from "@ethersproject/bignumber"
 import ERC20_ABI from "../constants/abis/erc20.json"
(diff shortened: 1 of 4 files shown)


src/providers/BasicPoolsProvider.tsx
```diff
@@ -33,54 +32,61 @@ import META_SWAP_ABI from "../constants/abis/metaSwap.json"
...
-  isSaddleApproved: boolean
-  isRemoved: boolean
-  isGuarded: boolean
-  isPaused: boolean
-  virtualPrice: BigNumber
   adminFee: BigNumber
-  swapFee: BigNumber
   aParameter: BigNumber
-  tokenBalances: BigNumber[]
+  isGuarded: boolean
```
hammeiam, line 44: just sorted these for easier scanning

src/providers/BasicPoolsProvider.tsx
```diff
@@ -33,54 +32,61 @@ import META_SWAP_ABI from "../constants/abis/metaSwap.json"
...
   basePoolAddress: null
+  isMetaSwap: false
   metaSwapDepositAddress: null
   underlyingTokenBalances: null
-  isMetaSwap: false
+  underlyingTokens: null
 }
 
 type SwapInfo = MetaSwapInfo | NonMetaSwapInfo
+type PoolMigrationData =
```
hammeiam, line 79: added more specific type for migration, corrected attribute name.

src/providers/BasicPoolsProvider.tsx
```diff
@@ -166,6 +172,10 @@ export async function getPoolsDataFromRegistry(
   poolRegistryMultiCall: MulticallContract<PoolRegistry>,
   ethCallProvider: MulticallProvider,
 ): Promise<SwapInfo[]> {
+  // curried fn to avoid passing in redundant arguments
+  function chunkedTryAll20<T>(calls: MulticallCall<unknown, T>[]) {
```
hammeiam, line 176: helper function. 20 may not be the right number.

src/providers/BasicPoolsProvider.tsx
```diff
@@ -176,75 +186,43 @@ export async function getPoolsDataFromRegistry(
...
-        aParametersMultiCalls.push(poolRegistryMultiCall.getA(poolAddress))
-        tokenBalancesMultiCalls.push(
-          poolRegistryMultiCall.getTokenBalances(poolAddress),
-        )
-        underlyingTokenBalancesMultiCalls.push(
-          poolRegistryMultiCall.getUnderlyingTokenBalances(poolAddress),
-        )
-      }
+    const poolMulticalls = (
+      registryPoolData.filter(Boolean) as {
```
hammeiam, line 190: doing `filter(Boolean)` lets us skip the explicit null check, and also use map instead of `forEach` or `reduce`

src/providers/BasicPoolsProvider.tsx
```diff
@@ -176,75 +186,43 @@ export async function getPoolsDataFromRegistry(
...
+      )
+      return [
+        lpTokenContract.totalSupply(),
+        poolRegistryMultiCall.getA(poolAddress),
+        poolRegistryMultiCall.getPaused(poolAddress),
+        poolRegistryMultiCall.getSwapStorage(poolAddress),
+        poolRegistryMultiCall.getTokenBalances(poolAddress),
+        poolRegistryMultiCall.getUnderlyingTokenBalances(poolAddress),
+        poolRegistryMultiCall.getVirtualPrice(poolAddress),
+      ] as const
```
hammeiam, line 207: using `as const` allows us to keep the exact types of each item in the tuple (no longer seen as an array)

hammeiam: also sorted by name for easier scanning

src/providers/BasicPoolsProvider.tsx
```diff
@@ -176,75 +186,43 @@ export async function getPoolsDataFromRegistry(
...
-      ethCallProvider.tryAll(swapStoragesMulticalls),
-      ethCallProvider.tryAll(aParametersMultiCalls),
-      ethCallProvider.tryAll(tokenBalancesMultiCalls),
-      chunkedTryAll<BigNumber[]>(
-        underlyingTokenBalancesMultiCalls,
-        ethCallProvider,
-        2,
-      ),
-      ethCallProvider.tryAll(lpTokenTotalSuppliesMultiCalls),
+      chunkedTryAll20(poolMulticalls.map((a) => a[0])), // lpTokenSupply
```
hammeiam, line 219: The one caveat to this approach is that it's important to get the indexes correct.

src/providers/BasicPoolsProvider.tsx
```diff
@@ -256,55 +234,55 @@ export async function getPoolsDataFromRegistry(
           chainId,
           poolData.poolAddress.toLowerCase(),
         )
-        const isSynthetic = (
-          mapToLowerCase(poolData.tokens) ||
-          underlyingTokensBalances[index] ||
```
hammeiam, line 261: using underlying balances was a bug

src/providers/BasicPoolsProvider.tsx
```diff
@@ -256,55 +234,55 @@ export async function getPoolsDataFromRegistry(
...
+        ).some((addr) => isSynthAsset(chainId, addr))
         const isPaused = arePoolsPaused[index]
         const virtualPrice = virtualPrices[index]
         const swapStorage = swapStorages[index]
         const aParameter = aParameters[index]
         const tokenBalances = tokensBalances[index]
-        const underlyingTokenBalances = underlyingTokensBalances[index] || []
+        const underlyingTokenBalances =
+          underlyingTokensBalances[index] ||
+          Array(poolData.underlyingTokens.length).fill(Zero)
```
hammeiam, line 247: fill with zero if the call failed.

src/providers/BasicPoolsProvider.tsx
```diff
@@ -256,55 +234,55 @@ export async function getPoolsDataFromRegistry(
...
         const lpTokenSupply = lpTokensSupplies[index]
 
         if (
-          isPaused == null ||
-          virtualPrice == null ||
           aParameter == null ||
+          isPaused == null ||
```
hammeiam, line 252: sorted

src/providers/BasicPoolsProvider.tsx
```diff
@@ -256,55 +234,55 @@ export async function getPoolsDataFromRegistry(
...
-          lpTokenSupply == null
+          virtualPrice == null
         ) {
           return
         }
 
-        const sharedSwapData: SharedSwapData = {
-          poolAddress: poolData.poolAddress.toLowerCase(),
+        const sharedSwapData = {
+          adminFee: swapStorage.adminFee,
```
hammeiam, line 262: sorted

src/providers/BasicPoolsProvider.tsx
```diff
@@ -321,12 +299,12 @@ export async function getPoolsDataFromRegistry(
 }
 
 function buildMetaInfo(
-  sharedSwapData: DeepNullable<SharedSwapData>,
```
hammeiam, line 324: seems like we don't actually need DeepNullable, or at least we don't need it here since you already did null checks before this fn

src/utils/migrations.ts
```diff
@@ -35,14 +34,11 @@ export async function getMigrationData(
     const calls = poolAddresses.map((address) =>
       migratorContract.migrationMap(address),
     )
-    const results = await ethCallProvider.tryEach(
-      calls,
-      Array(calls.length).fill(true) as true[],
-    )
+    const results = await ethCallProvider.tryAll(calls)
```
hammeiam, line 37: update to the more modern tryAll pattern

src/providers/BasicPoolsProvider.tsx
```diff
@@ -33,54 +32,61 @@ import META_SWAP_ABI from "../constants/abis/metaSwap.json"
...
+  poolName: string
+  swapFee: BigNumber
+  targetAddress: string | null // the address from which a pool was cloned
+  tokenBalances: BigNumber[]
+  tokens: string[]
+  typeOfAsset: PoolTypes
+  virtualPrice: BigNumber
 }
 
 type MetaSwapInfo = SharedSwapData & {
```
hammeiam, line 62: tbh I liked keeping `SharedSwapData` as only the types that were shared, so I changed it back

hammeiam:
Closing as Duy has merged my changes to his branch

PR #1074, opened 2022-06-09
penalty percent fix

nkrulikovsky:
## What does this PR do?

- Insert votingEscrowContract address for mainnet so that we can test after `npm run fork:mainnet`
- Fix penalty percent 
- Fix penalty Alert

## Screenshot
Screen Shot 2022-06-09 at 9 23 40 PM
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           cache: "npm"
       # fix for forcing git to use https when pulling deps
       - run: 'git config --global --replace-all url."https://github.com/".insteadOf ssh://git@github.com/'
-      - run: npm ci --silent
+      - run: npm ci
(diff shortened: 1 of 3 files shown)


ug02fast (approved):

hammeiam:
I think your numbers are reversed, the modal in your screenshot says you're receiving 97% of your locked SDL, but I think it should be `1 - penaltyPercent` or 3% in this case

nkrulikovsky:
> I think your numbers are reversed, the modal in your screenshot says you're receiving 97% of your locked SDL, but I think it should be `1 - penaltyPercent` or 3% in this case

My mistake 😞 

Because on the alert we are having xxx penalty, so I thought penalty is the amount which user will lose.

Fixed as you suggested.

PR #1090, opened 2022-06-18
Add master_registry contract address on evmos

nkrulikovsky:


ug02fast (approved):
Is this why evmos is having problems?

nkrulikovsky:
@ug02fast  This PR is fixing the issue which saddle pool don't load on the frontend on the network other than `mainnet`.
But I can find the some lagging on `emvos` and in some case, it doesn't all their pool. I think we can fix that issue on another PR.

ug02fast (approved):

PR #1100, opened 2022-06-21
My Gauge Farm conditional UI for LP staked in minichef contracts

ug02fast:
This is for the case where user has existing LP staked in minichef and potentially also has stake in gauges.

image


hammeiam (approved):
Could the "old" unstake button say "Unstake Old" instead of "unstake all"? Previously "all" meant all lp tokens, but now that there are multiple staking contracts, it could be construed as "unstake from all contracts". Maybe also make the "unstake old" button the ternary orange or something to differentiate it.  @sandrasong

PR #1107, opened 2022-06-21
Add SDL icon & price to header

nguyenaiden:
***Description***: This is to add a button at the top Menu nav bar to include a SDL price tracker. SDL price information will render once CoinGecko starts to return a value for the SDL token. This work also include updating the geckoId of SDL token from `saddle-dao` to `saddle-finance`
#1103 
image

### DIFF
--- a/src/components/TopMenu.tsx
+++ b/src/components/TopMenu.tsx
@@ -1,3 +1,5 @@
+//@ts-nocheck
+// @eslint-ignore
 import {
   AppBar,
   Box,
@@ -13,6 +15,7 @@ import { IS_SDL_LIVE, IS_VESDL_LIVE } from "../constants"
 import { NavLink, NavLinkProps, useLocation } from "react-router-dom"
 import React, { ReactElement, useContext, useState } from "react"
 
+import { AppState } from "../state"
 import { IS_DEVELOPMENT } from "../utils/environment"
 import { MoreVert } from "@mui/icons-material"
 import NetworkDisplay from "./NetworkDisplay"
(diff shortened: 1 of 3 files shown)


src/components/TopMenu.tsx
```diff
@@ -44,7 +46,8 @@ function TopMenu(): ReactElement {
   const theme = useTheme()
   const { pathname } = useLocation()
   const activeTab = pathname.split("/")[1] as ActiveTabType
-
+  const { tokenPricesUSD } = useSelector((state: AppState) => state.application)
+  const sdlPrice = tokenPricesUSD?.[SDL_TOKEN.geckoId]
```
hammeiam, line 50: tokenPrices are keyed by symbol. so you can just use `"SDL"` here

src/components/TopMenu.tsx
```diff
@@ -176,4 +180,30 @@ function RewardsButton({
...
+      <Button
+        variant="contained"
+        color="info"
+        data-testid="sdlPriceButton"
+        href={SUSHI_WETH_SDL_POOL_URL}
+        target="_blank"
+        startIcon={<SaddleLogo width={20} height={20} />}
+        style={{ minWidth: 250 }}
+      >
+        {`SDL Price: ${sdlPrice}`}
```
hammeiam: Since this takes up a lot of space, maybe we can just format it as `$0.25` (dollar sign, no text, 2 decimal places) since we already have the icon there

hammeiam (approved):

src/components/TopMenu.tsx
nguyenaiden: I wanted to use what's already there in the constants file.

ug02fast (approved):

PR #1108, opened 2022-06-22
feat: get explorer link by chainid

0xEapp:
currently when a user wants to check transactions records only etherscan is avaible, although a user may be on fantom chain for example the link redirects to ehterscan. so the solution here is to utilize the getMultiChainScanLink function and render the links on front end by chainId
Before Screen shot:
before-screenshot

After Screenshot:
after-screenshot


ug02fast (approved):
Lgtm

PR #1114, opened 2022-06-22
Vesdl calculator fix

nkrulikovsky:
- Remove the pools which outdated because these pools don't have gauge.


src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -122,15 +122,18 @@ export default function VeTokenCalculator({
...
-            </MenuItem>
             {basicPools &&
-              Object.keys(basicPools)?.map((poolName) => (
-                <MenuItem key={poolName} value={poolName}>
-                  {poolName}
-                </MenuItem>
-              ))}
+              Object.keys(basicPools)?.map((poolName) => {
+                const pool = basicPools[poolName]
+                const outdated =
```
ug02fast, line 128: instead of `return null` can you use .filter before the `.map` to filter out entries that don't belong?

src/pages/VeSDL/VeTokenCalculator.tsx
```diff
@@ -122,15 +122,18 @@ export default function VeTokenCalculator({
...
+                  pool?.isGuarded || pool?.isMigrated || pool?.isPaused
+                if (outdated) return null
+                return (
+                  <MenuItem key={poolName} value={poolName}>
```
ug02fast, line 132: poolNames unique?

ug02fast (commented):
looks good if David approves too

src/pages/VeSDL/VeTokenCalculator.tsx
nkrulikovsky: OK, Good.

src/pages/VeSDL/VeTokenCalculator.tsx
nkrulikovsky: We can use pool address instead of pool name.

nkrulikovsky: Udated!

ug02fast (approved):

PR #1118, opened 2022-06-23
Nikita/farm token fix

nkrulikovsky:
## Farm token fix.
Farm token doesn't display because in some case pull name from gauge is null.

It fix this issue by getting poolTokens from basicPoolContext.

But the following 3 gauges don't have pool addresss.

`saddleTBTC-V3-gauge`
`saddleWCUSD-V3-gauge`
`saddleUSDCUSDTFrax-gauge`

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -1859,7 +1859,7 @@ export const SYNTH_TRACKING_ID =
 export const IS_VIRTUAL_SWAP_ACTIVE = true
 export const IS_L2_SUPPORTED = true
 export const IS_SDL_LIVE = true
-export const IS_VESDL_LIVE = true
+export const IS_VESDL_LIVE = false
 export const IS_POOL_REGISTRY_MIGRATION_LIVE = false
 // FLAGS END
 
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -1859,7 +1859,7 @@ export const SYNTH_TRACKING_ID =
 export const IS_VIRTUAL_SWAP_ACTIVE = true
 export const IS_L2_SUPPORTED = true
 export const IS_SDL_LIVE = true
-export const IS_VESDL_LIVE = false
+export const IS_VESDL_LIVE = true
 export const IS_POOL_REGISTRY_MIGRATION_LIVE = false
 // FLAGS END
 
--- a/src/pages/Farm/FarmOverview.tsx
+++ b/src/pages/Farm/FarmOverview.tsx
@@ -1,13 +1,20 @@
 import { Box, Button, Grid, Typography, styled } from "@mui/material"
-import { formatBNToPercentString, formatBNToShortString } from "../../utils"
+import {
+  commify,
+  formatBNToPercentString,
+  formatBNToShortString,
+  formatBNToString,
+} from "../../utils"
+
 import { BigNumber } from "@ethersproject/bignumber"
+import { GaugeApr } from "../../providers/AprsProvider"
 import React from "react"
 import TokenIcon from "../../components/TokenIcon"
 import { useTranslation } from "react-i18next"
 
 interface FarmOverviewProps {
   farmName: string
-  apr?: BigNumber
+  aprs?: GaugeApr[]
   tvl?: BigNumber
   myStake?: BigNumber
   onClickStake: () => void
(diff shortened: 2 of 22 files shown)


src/pages/Farm/FarmOverview.tsx
```diff
@@ -61,13 +59,18 @@ FarmOverviewProps): JSX.Element {
...
-              <TokenIcon
-                key={token.symbol}
-                symbol={token.symbol}
-                alt={token.symbol}
-              />
-            ))
+            poolTokens?.map((tokenAddress) => {
+              if (!chainId) return <div></div>
+              const token = getTokenByAddress(tokenAddress, chainId)
+              if (!token) return <div></div>
```
ug02fast, line 65: If it makes sense, instead of returning <div></div> can you `.filter` with same condition before the `.map`?

src/pages/Farm/FarmOverview.tsx
```diff
@@ -61,13 +59,18 @@ FarmOverviewProps): JSX.Element {
...
+              return (
+                <TokenIcon
+                  key={token?.name}
+                  symbol={token.symbol}
+                  alt={token?.symbol}
```
ug02fast, line 70: Can you change these to be of a similar pattern?

src/pages/Farm/FarmOverview.tsx
```diff
@@ -61,13 +59,18 @@ FarmOverviewProps): JSX.Element {
...
           ) : (
-            farmTokens.map((token) => (
-              <TokenIcon
-                key={token.symbol}
-                symbol={token.symbol}
-                alt={token.symbol}
-              />
-            ))
+            poolTokens?.map((tokenAddress) => {
+              if (!chainId) return <div></div>
```
ug02fast, line 63: Possible move and handle this `!chainId` check higher up? Outside of the loop?

ug02fast (changes requested):

hammeiam (approved):
Not ideal, but it works! We'll need to remove getTokenByAddress in the future. I'm still not sure why usePoolData wasn't working for this

src/pages/Farm/FarmOverview.tsx
nkrulikovsky: Because I think we don't able to use filter because the poolTokens is only address of token, not token.

src/pages/Farm/FarmOverview.tsx
nkrulikovsky: good call out!

src/constants/index.ts
```diff
@@ -1859,7 +1859,7 @@ export const SYNTH_TRACKING_ID =
 export const IS_VIRTUAL_SWAP_ACTIVE = true
 export const IS_L2_SUPPORTED = true
 export const IS_SDL_LIVE = true
-export const IS_VESDL_LIVE = false
+export const IS_VESDL_LIVE = true
```
hammeiam, line 1862: DO NOT MERGE THIS

hammeiam (changes requested):

hammeiam (approved):

ug02fast (approved):
lgtm, you can also return `null` if chainId falsy

src/pages/Farm/FarmOverview.tsx
ug02fast:
```
poolTokens?.filter((tokenAddress) => Boolean(getTokenByAddress(tokenAddress, chainId)).map((tokenAddress) =>  
{
  const token = getTokenByAddress(...
 ```
 
Something like this should work but you have to double-check if you want to do this. After filtering you will only have valid tokens so there will be no use for `if (!token)` case.

File: src/pages/Farm/FarmOverview.tsx [hunk 5]
```diff
@@ -29,17 +29,16 @@ const TokenGroup = styled("div")(() => ({
...
   myStake,
   onClickStake,
 }: // onClickClaim,
 FarmOverviewProps): JSX.Element {
   const { t } = useTranslation()
-  const [farmData] = usePoolData(poolName || "")
-  const farmTokens = farmData.tokens
-
+  const { chainId } = useActiveWeb3React()
+  if (!chainId) return <div></div>
```
ug02fast, line 41: Can you add extra newline after this for style purposes? You can also return `null`

nkrulikovsky: BTW, getTokenByAddress is return `Token||null` so after filtering we will get null again so we need to check again.

PR #1119, opened 2022-06-23
Nikita/vesdl unlock helper fix

nkrulikovsky:
Don't display helper text when it is expired.

### DIFF
--- a/src/pages/VeSDL/index.tsx
+++ b/src/pages/VeSDL/index.tsx
@@ -157,14 +157,15 @@ export default function VeSDL(): JSX.Element {
   // Calculate penalty Ratio and penalty amount
   const leftTimeForUnlock = lockEnd && getUnixTime(lockEnd) - currentTimestamp
   const isExpired = leftTimeForUnlock && leftTimeForUnlock <= 0
-  const penaltyAmount = !isExpired
-    ? minBigNumber(
-        lockedSDLVal.mul(BigNumber.from(3)).div(BigNumber.from(4)),
-        lockedSDLVal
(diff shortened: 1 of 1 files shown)


ug02fast (approved):

src/pages/VeSDL/index.tsx
```diff
@@ -468,7 +470,7 @@ export default function VeSDL(): JSX.Element {
                 )}
               </Typography>
             </Typography>
-            {!penaltyAmount.isZero() && leftTimeForUnlock && (
+            {!penaltyAmount.isZero() && leftTimeForUnlock && !isExpired && (
```
hammeiam, line 473: These all mean the same thing, correct? If `isExpired`, then `leftTimeForUnlock <= 0`, and `penaltyAmount <= 0`. This can be simplified.

hammeiam (approved):
approving to unblock

src/pages/VeSDL/index.tsx
hammeiam: Also `penaltyAmount` may be less than zero, so `penaltyAmount.isZero()` would be false. Better to use `penaltyAmount.lte(Zero)`

nkrulikovsky: Yes, but because of type script, if we don't check the leftTimeForUnlock, `secondsToHours(leftTimeForUnlock)` makes error even though we are check it on `!isExpired`.
That is the reason why I repeated same logic again.

And I agree to use `penaltyAmout.lte(Zero)`

PR #1127, opened 2022-06-23
Display boost multiplier on StakingDialog modal

nguyenaiden:
***Description***: On the staking dialog modal, user's current boost multiplier will be calculated and displayed. If the user hasn't provided LP or staked any asset, the multiplier will show up as `-`
image
image

### DIFF
--- a/src/hooks/useUserGauge.ts
+++ b/src/hooks/useUserGauge.ts
@@ -58,10 +58,7 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
   )
 
   const boost =
-    gauge.gaugeBalance != null &&
-    !gauge.gaugeBalance.isZero() &&
-    gauge.workingBalances != null &&
-    !gauge.workingBalances.isZero()
+    gauge.gaugeBalance?.gt(Zero) && gauge.workingBalances?.gt(Zero)
       ? `${formatUnits(
           gauge.workingBalances
             .mul(BN_1E18)
--- a/src/hooks/useUserGauge.ts
+++ b/src/hooks/useUserGauge.ts
@@ -4,6 +4,7 @@ import {
   useLiquidityGaugeContract,
 } from "./useContract"
 
+import { BN_1E18 } from "../constants"
 import { BigNumber } from "@ethersproject/bignumber"
 import { ContractTransaction } from "ethers"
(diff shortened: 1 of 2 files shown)


src/hooks/useUserGauge.ts
```diff
@@ -54,6 +57,21 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
     userGaugeRewards?.claimableExternalRewards.length,
   )
 
+  const boost =
+    gauge.gaugeBalance != null &&
+    !gauge.gaugeBalance.isZero() &&
+    gauge.workingBalances != null &&
+    !gauge.workingBalances.isZero()
```
ug02fast, line 64:
```
gauge?.gaugeBalance?.gt(Zero) &&
gauge?.workingBalances?.gt(Zero)
```
this should accomplish similar if you wanted to simplify

ug02fast (approved):

src/hooks/useUserGauge.ts
ug02fast: might just be 
```
gauge.gaugeBalance?.gt(Zero) &&
gauge.workingBalances?.gt(Zero)
```

nguyenaiden: Good call. Will use it now.

src/hooks/useUserGauge.ts
```diff
@@ -54,6 +57,18 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
     userGaugeRewards?.claimableExternalRewards.length,
   )
 
+  const boost =
+    gauge.gaugeBalance?.gt(Zero) && gauge.workingBalances?.gt(Zero)
+      ? `${formatUnits(
+          gauge.workingBalances
+            .mul(BN_1E18)
+            .div(gauge.gaugeBalance)
```
hammeiam, line 65: Careful when dividing by unknown values - they may be 0 and BigNumber will throw

hammeiam (approved):
Looks good besides one gotcha, approving to unblock

src/hooks/useUserGauge.ts
nguyenaiden: we have a null and non Zero check above.

PR #1128, opened 2022-06-23
Allow claiming of minichef rewards in token dialog

ug02fast:
This will render minichef "outdated" lp rewards for the user to claim if there are pools with rewards to claim. This was not accounted for initially.


nguyenaiden (approved):
+1 to unblock. We'll go back and fix the CI issue post launch.

PR #1131, opened 2022-06-24
VeSDL calculator fix

nkrulikovsky:
- no vesdl
image

image

- After locking max vesdl
image


nguyenaiden (approved):
Looks good! I'll leave the CSS handling of the `veSDL for Max Boost` that's being wrapped due to the value being too large up to you

ug02fast (approved):

PR #1143, opened 2022-06-27
Add sdl lock time

hammeiam:
Avg SDL lock time = (veSDL totalSupply / SDL locked in veSDL) * max lock duration

Screen Shot 2022-06-27 at 10 26 35 AM


src/pages/VeSDL/LockedInfo.tsx
```diff
@@ -64,10 +70,16 @@ export default function LockedInfo(): JSX.Element {
           {aggSDLInfo ? formatBNToShortString(aggSDLInfo.totalVeSDL, 18) : "-"}
         </Typography>
       </Box>
-      {/* <Box flex={1}>
```
ug02fast, line 67: You want to preserve this for the future?

hammeiam: we're in the future

PR #1144, opened 2022-06-27
Connect wallet instructions when wallet not connected.

ug02fast:
image
Instruction to connect wallet until we can find a solution for loading pools without a connected account.

image
image


nkrulikovsky (approved):
Should we use this band on the farm page ?

ug02fast:
> Should we use this band on the farm page ?

ye

PR #1161, opened 2022-07-01
Fix staking max input

hammeiam:
TokenInput `max` value needs to be an unformatted string - will break if given something with more than one comma and/or period. 
Verified that this is the only instance of `max` being formatted


nkrulikovsky (approved):

nkrulikovsky:
Don't merge

PR #1162, opened 2022-07-01
Fetch isKIlled info for gauges

hammeiam:


ug02fast (approved):

src/components/GaugeWeight.tsx
```diff
@@ -26,10 +26,11 @@ export default function GaugeWeight({
   const theme = useTheme()
 
   const gaugesInfo = Object.values(gauges)
-    .map(({ gaugeName, gaugeRelativeWeight, poolAddress }) => {
+    .map(({ gaugeName, gaugeRelativeWeight, poolAddress, isKilled }) => {
       const pool = Object.values(basicPools || {}).find(
```
nguyenaiden, line 30: I think we can move the `isKilled` up here or we can just do a `.filter` before or after the `.map`

PR #1167, opened 2022-07-05
Fix SLP fetching issue

hammeiam:
For some reason, the codebase had issues fetching information about the SLP pair, causing the TVL of the pair not to render. This was resolved by remove the reference to account from `useSdlWethSushiPairContract`. I removed it from a few other places as well for good housekeeping.

I also update the sushi link from the analytics page to https://ethereum.sushi.com/swap?inputCurrency=ETH&outputCurrency=0xf1Dc500FdE233A4055e25e5BbF516372BC4F6871


hammeiam:
> Why does removing reference to account solve the issue?

🤷‍♂️

PR #1169, opened 2022-07-05
useApproveAndDeposit POOLS_MAP removed

ug02fast:
useApproveAndDeposit POOLS_MAP removed

### DIFF
--- a/src/hooks/useApproveAndDeposit.ts
+++ b/src/hooks/useApproveAndDeposit.ts
@@ -113,7 +113,10 @@ export function useApproveAndDeposit(
         token: BasicToken | undefined,
       ): Promise<void> => {
         if (!token || !library) {
-          enqueueToast("error", "Token or library is not loaded")
+          enqueueToast(
+            "error",
+            "There was a problem loading the token or library",
+          )
           throw new Error("Token or library is not loaded")
         }
         const spendingValue = BigNumber.from(state[token.symbol].valueSafe)
--- a/src/hooks/useApproveAndDeposit.ts
+++ b/src/hooks/useApproveAndDeposit.ts
@@ -112,7 +112,10 @@ export function useApproveAndDeposit(
       const approveSingleToken = async (
         token: BasicToken | undefined,
       ): Promise<void> => {
-        if (!token || !library) return
+        if (!token || !library) {
+          enqueueToast("error", "Token or library is not loaded")
+          throw new Error("Token or library is not loaded")
+        }
         const spendingValue = BigNumber.from(state[token.symbol].valueSafe)
         if (spendingValue.isZero()) return
         const tokenContract = getContract(
--- a/src/hooks/useApproveAndDeposit.ts
+++ b/src/hooks/useApproveAndDeposit.ts
@@ -1,11 +1,10 @@
 import {
   BTC_POOL_NAME,
-  POOLS_MAP,
   PoolName,
   TRANSACTION_TYPES,
-  Token,
   isLegacySwapABIPool,
 } from "../constants"
+import { BasicToken, TokensContext } from "../providers/TokensProvider"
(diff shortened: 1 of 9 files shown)


src/hooks/useApproveAndDeposit.ts
```diff
@@ -61,17 +62,19 @@ export function useApproveAndDeposit(
...
   } = useSelector((state: AppState) => state.user)
   const POOL = POOLS_MAP[poolName]
+  const basicPools = useContext(BasicPoolsContext)
+  const pool = basicPools?.[poolName]
   const metaSwapContract = useMemo(() => {
-    if (POOL.metaSwapAddresses && chainId && library) {
+    if (pool?.metaSwapDepositAddress && chainId && library) {
       return getContract(
-        POOL.metaSwapAddresses?.[chainId],
+        pool.metaSwapDepositAddress[chainId],
```
ug02fast, line 70: @hammeiam can you confirm this naming conversion is correct? I read the comment about it but want to make sure.

src/hooks/useApproveAndDeposit.ts
```diff
@@ -60,18 +60,20 @@ export function useApproveAndDeposit(
     transactionDeadlineSelected,
     infiniteApproval,
   } = useSelector((state: AppState) => state.user)
-  const POOL = POOLS_MAP[poolName]
+  const basicPools = useContext(BasicPoolsContext)
+  const tokens = useContext(TokensContext)
+  const pool = basicPools?.[poolName]
   const metaSwapContract = useMemo(() => {
-    if (POOL.metaSwapAddresses && chainId && library) {
+    if (pool?.poolAddress && chainId && library) {
```
ug02fast, line 67: @hammeiam can you confirm this naming conversion is correct? I read the comment about it but want to make sure.

ug02fast:
> Mostly minor feedback about doing null checks at the start of functions to avoid a lot of option chaining later on. Logic looks good.
> 
> Double check that you can correctly approve and deposit on localhost.
> 
> I'd also be curious to get your thoughts on how you'd refactor this file, since this is one of the oldest in the repo and could definitely be improved.
> 
> Approving to unblock.

I confirmed approval and deposit works on localhost.

`approveSingleToken` can be pulled out into a hook utility file.
Gas logic can be deleted as it's not utilized.
Some of the conditional statements could be refactored to be simpler. 
Deadline and approval logic can be refactored into a function and pulled out of the file.
metaswapContract can be pulled out of the file and imported.

src/hooks/useApproveAndDeposit.ts
```diff
@@ -107,10 +109,18 @@ export function useApproveAndDeposit(
         gasPriceUnsafe ? String(gasPriceUnsafe) : "45",
         9,
       )
-      const approveSingleToken = async (token: Token): Promise<void> => {
+      const approveSingleToken = async (
+        token: BasicToken | undefined,
+      ): Promise<void> => {
+        if (!token || !library) return
```
hammeiam, line 115: Would prefer to give the user a visible error here since this fn is called when the user clicks a button.

src/hooks/useApproveAndDeposit.ts
```diff
@@ -113,7 +113,10 @@ export function useApproveAndDeposit(
         token: BasicToken | undefined,
       ): Promise<void> => {
         if (!token || !library) {
-          enqueueToast("error", "Token or library is not loaded")
+          enqueueToast(
+            "error",
+            "There was a problem loading the token or library",
+          )
           throw new Error("Token or library is not loaded")
```
hammeiam, line 120: I think throwing here will result in an uncaught error - we can just console.error instead. I think generally we should have ~2 types of error messages: one for the user eg "There was a problem with your transaction" displayed as a UI notification, and one for debugging or technical users eg "library is undefined" + stack trace, logged to the console and Sentry.

PR #1176, opened 2022-07-11
Allow  Wrapped Token withdraw

nkrulikovsky:
## What does this PR do?

enable to withdraw wrapped token from meta pool.

NOTE
Withdrawing function is having more issue like the following.
- cannot withdraw one token
- max value don't display over the TokenInput ( we are giving Zero to underlying token on meta pool )


src/hooks/useApproveAndWithdraw.ts
```diff
@@ -33,6 +36,19 @@ export function useApproveAndWithdraw(
   const { gasStandard, gasFast, gasInstant } = useSelector(
     (state: AppState) => state.application,
   )
+
+  const POOL = POOLS_MAP[poolName]
```
ug02fast, line 40: Can you use basicPoolContext pools here?

src/hooks/useWithdrawFormState.ts
```diff
@@ -37,17 +37,16 @@ export type WithdrawFormAction = {
   fieldName: FormFields | "reset"
   address?: string
   value: string
+  max?: BigNumber //TODO: get underlying token balance from registry pool
```
ug02fast, line 40: Can you use registry pool and get the underlying token balance?

ug02fast (changes requested):
Nice work. Can you leverage the registry pools now instead of POOLS_MAP?

src/hooks/useApproveAndWithdraw.ts
nkrulikovsky: Good callout!

src/hooks/useWithdrawFormState.ts
nkrulikovsky: On `BasicPoolsProvider.tsx`, we are getting underlyingTokenBalances on line 249.

ug02fast: great, is there a problem with using pools from registry then?

ug02fast: Should the TODO: comment still be there?

nkrulikovsky: Yes, on line 112, we are getting pools from getPoolsBaseData on BasicPoolsProvider.
So, underlyingTokenBalance is Zero now.

PR #1177, opened 2022-07-11
Fix gauge rewards claiming logic

hammeiam:
Now explicitly check that that we have minter rewards before creating minter txn, check that we have gauge rewards before creating gauge claim txn. Give more explicit feedback to user.

### DIFF
--- a/src/components/TokenClaimDialog.tsx
+++ b/src/components/TokenClaimDialog.tsx
@@ -468,8 +468,8 @@ function useRewardClaims() {
       try {
         updateClaimStatus(gauge.gaugeName, STATUSES.PENDING)
         const claimPromises = []
-        const minterRewards = gauge.rewards.filter(({ isSDL }) => isSDL)
-        const gaugeRewards = gauge.rewards.filter(({ isSDL }) => !isSDL)
+        const minterRewards = gauge.rewards.filter(({ isMinter }) => isMinter)
+        const gaugeRewards = gauge.rewards.filter(({ isMinter }) => !isMinter)
         if (minterRewards.length > 0) {
           claimPromises.push(gaugeMinterContract.mint(gauge.address))
         }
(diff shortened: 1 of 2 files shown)


ug02fast (approved):

src/components/TokenClaimDialog.tsx
```diff
@@ -453,12 +453,24 @@ function useRewardClaims() {
...
       } | null,
     ) => {
-      if (!chainId || !account || !gaugeMinterContract || !library || !gauge)
+      if (!chainId || !account || !gaugeMinterContract || !library || !gauge) {
+        enqueueToast("error", "Unable to claim reward")
+        return
+      }
+
+      if (gauge.rewards.length === 0) {
+        enqueueToast("error", "No rewards to claim")
```
nguyenaiden, line 462: do we want to enqueue an error toast if there's no rewards available? Can we just gray out the claim button instead?

hammeiam: ack

hammeiam: I like including this in the claim hook as well, just in case the UI doesn't disable. This should be an edge case.

nguyenaiden: sounds good!

src/components/TokenClaimDialog.tsx
```diff
@@ -453,12 +456,24 @@ function useRewardClaims() {
...
         return
+      }
       try {
         updateClaimStatus(gauge.gaugeName, STATUSES.PENDING)
         const claimPromises = []
-        if (gauge.rewards.length > 0 && gauge.address) {
+        const minterRewards = gauge.rewards.filter(({ isSDL }) => isSDL)
```
nguyenaiden, line 471: Can we update the var names to `sdlRewards` and `externalRewards` to make things clearer? Anything helps since the entire concept is pretty confusing anyway.

PR #1182, opened 2022-07-13
Registry migration swap refactor

ug02fast:
Refactored POOLS_MAP out of Swap components. Tested swap successfully in hardhat enviornment.

Mostly renames and early returns after adjusting [registryMigrationTokensMapRefactor](https://github.com/saddle-finance/saddle-frontend/tree/registryMigrationTokensMapRefactor)


nguyenaiden (approved):
This is solid to get this migration in a good place. We should implement default states to reduce null checks and null coalescing throughout the codebase later.

src/hooks/useTokenMaps.ts
```diff
@@ -0,0 +1,43 @@
...
+    })
+    const lpTokenSymbol = tokens?.[pool?.lpToken ?? ""]?.symbol ?? ""
+    const lpToken = tokens?.[pool?.lpToken ?? ""]
+    newAcc[lpTokenSymbol] = lpToken
+    return newAcc
+  }, {} as BasicTokensMap)
+
+  const tokenToPoolsMap = Object.keys(basicPools ?? {}).reduce(
+    (acc, poolName) => {
+      const pool = basicPools?.[poolName]
```
hammeiam, line 29: guard pattern after this

src/hooks/useTokenMaps.ts
```diff
@@ -0,0 +1,43 @@
...
+      const poolTokens = pool?.tokens.map((token) => tokens?.[token]) ?? []
+      const newAcc = { ...acc }
+      poolTokens.forEach((token) => {
+        newAcc[token?.symbol ?? ""] = (
+          newAcc[token?.symbol ?? ""] || []
+        ).concat(poolName as PoolName)
```
hammeiam, line 35: PoolName also needs to be killed as part of the move from POOL_MAP

src/hooks/useTokenMaps.ts
```diff
@@ -0,0 +1,43 @@
+import { BasicTokensMap, PoolName, TokenToPoolsMap } from "../constants"
+
+import { BasicPoolsContext } from "../providers/BasicPoolsProvider"
+import { TokensContext } from "../providers/TokensProvider"
+import { useContext } from "react"
+
+export const useTokenMaps = (): {
```
hammeiam, line 7: Should this also account for `underlyingTokens`, or just `tokens`?

hammeiam, line 7: Add a comment somewhere that this maps token symbols to various things, so devs don't try key on addresses or something else

src/pages/Swap.tsx
```diff
@@ -163,21 +162,21 @@ function Swap(): ReactElement {
...
             formState.currentSwapPairs
               .map(({ to, type: swapType }) => {
-                if (!TOKENS_MAP?.[to.symbol]) {
+                if (!tokensMap?.[to.symbol]) {
                   console.log("unknown symbol", { to, swapType })
                   return null
                 }
-                const { symbol, name, decimals } = TOKENS_MAP[to.symbol]
-                const amount = tokenBalances?.[symbol] || Zero
+                const token = tokensMap?.[to.symbol]
```
hammeiam, line 169: We know that `token` is defined based on the check at 165, thus this should be `const token = tokensMap[to.symbol] as BasicToken` and remove all of the `?? ""` below

src/pages/Swap.tsx
```diff
@@ -277,9 +282,9 @@ function Swap(): ReactElement {
         formStateArg.swapType === SWAP_TYPES.TOKEN_TO_SYNTH &&
         bridgeContract != null
       ) {
-        const originPool = POOLS_MAP[formStateArg.from.poolName]
+        const originPool = basicPools?.[formStateArg.from.poolName]
         amountToReceive = await bridgeContract.calcTokenToSynth(
-          originPool.addresses[chainId],
+          originPool?.metaSwapDepositAddress ?? "",
```
hammeiam, line 287: Defaulting to "" or 0 doesn't make sense for any of the values here and below - `calcTokenToSynth` will fail without a valid address, and all math will be wrong if we default a token's decimals to anything. If the pool or token is undefined, it's not even worth trying to execute the function that depends on those things.

hammeiam:
@nguyenaiden please take a closer look at PRs, I think there's a lot of good feedback that could've been added in your review

src/constants/index.ts
```diff
@@ -2060,6 +2061,10 @@ export type TokensMap = {
   [symbol: string]: Token
 }
 
+export type BasicTokensMap = {
```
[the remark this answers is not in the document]
ug02fast: Yeah, I even read somewhere the types should live in the respective component and not a single types file.

src/hooks/useCalculateSwapPairs.ts
```diff
@@ -99,11 +99,11 @@ export function useCalculateSwapPairs(): (token?: Token) => SwapData[] {
   }, [chainId])
 
   return useCallback(
-    function calculateSwapPairs(token?: Token): SwapData[] {
+    function calculateSwapPairs(token?: BasicToken): SwapData[] {
       if (token == null || chainId == null || tokens === null) return []
       const cacheHit = pairCache[token.symbol]
       if (cacheHit) return cacheHit
-      const originTokenAddress = (token.addresses[chainId] || "").toLowerCase()
+      const originTokenAddress = (token?.address || "").toLowerCase()
```
nguyenaiden, line 106: The token nullcheck already happened above. Don't think you need the optional here.

src/pages/Swap.tsx
```diff
@@ -128,11 +127,11 @@ function Swap(): ReactElement {
...
-      .filter(({ isLPToken, addresses }) => !isLPToken && addresses[chainId])
+    const allTokens = Object.values(tokens || {})
+      .filter(({ isLPToken, address }) => !isLPToken && address)
       .filter(({ symbol }) => {
         // get list of pools containing the token
-        const tokenPools = TOKEN_TO_POOLS_MAP[symbol]
+        const tokenPools = tokenToPoolsMap[symbol] ?? []
         // ensure at least one pool is unpaused to include token in swappable list
         const hasAnyUnpaused = tokenPools.some((poolName) => {
           const basicPool = basicPools?.[poolName]
```
nguyenaiden, line 137: Going forward, we should do a nullcheck and early exit if any context retrieval is null so we don't have optional chaining scattered everywhere. e.g: 
```
const basicPools = useContext(BasicPoolsContext) or const { tokensMap, tokenToPoolsMap } = useTokenMaps()
```

If we do a nullcheck at the top level for this, we can get rid of this optional on L237
```
tokensMap?.[formStateArg.from.symbol]?.decimals ?? 0,
```

nguyenaiden: Not part of your diff but on L209 of this file, convert ` if (tokenBalances === null || chainId == null)` to use `==` for both comparison for consistency.
And on L213 - L216, use a `[comparison value] == null` so that it checks for both nulls and undefined.

nguyenaiden (changes requested):
One of the comments I left re: nullchecks and early exit for context retrieval. We should set a standard for this so we don't have optionals everywhere. Thoughts @hammeiam?

src/hooks/useTokenMaps.ts
ug02fast: THis is a good catch. I believe it should be for underlyingTokens since they switched since bringing in PoolsRegistry.

src/pages/Swap.tsx
ug02fast: > Not part of your diff but on L209 of this file, convert ` if (tokenBalances === null || chainId == null)` to use `==` for both comparison for consistency. And on L213 - L216, use a `[comparison value] == null` so that it checks for both nulls and undefined.

This is ok but there are safer ways to protect against falsy values.

ug02fast: `!<comparison value>` which will check it for all falsy values with less syntax.

src/hooks/useTokenMaps.ts
nguyenaiden: @ug02fast  Bumping this. Make sure you have a comment here addressing this.

src/pages/Swap.tsx
```diff
@@ -162,21 +167,21 @@ function Swap(): ReactElement {
         ? (
             formState.currentSwapPairs
               .map(({ to, type: swapType }) => {
-                if (!tokensMap[to.symbol]) {
+                if (!tokenSymbolToTokenMap[to.symbol]) {
                   console.log("unknown symbol", { to, swapType })
```
nguyenaiden, line 171: This has been here, but was this intentional?

src/pages/Swap.tsx
```diff
@@ -131,21 +131,26 @@ function Swap(): ReactElement {
...
       .filter(({ symbol }) => {
         // get list of pools containing the token
-        const tokenPools = tokenToPoolsMap[symbol] ?? []
+        if (!tokenSymbolToPoolNameMap[symbol]) return false
+        const tokenPools = tokenSymbolToPoolNameMap[symbol]
         // ensure at least one pool is unpaused to include token in swappable list
         const hasAnyUnpaused = tokenPools.some((poolName) => {
-          const basicPool = basicPools?.[poolName]
-          return basicPool ? !basicPool.isPaused : false
+          if (!basicPools) return false
```
nguyenaiden, line 138: Could combine this to
``` 
if (basicPools == null || basicPools[poolName] == null) return false
return !basicPool[poolName].isPaused`
```
Since the right hand side won't run if the first condition is true.

src/pages/Swap.tsx
```diff
@@ -131,21 +131,26 @@ function Swap(): ReactElement {
...
+          if (!basicPool) return false
+          return !basicPool.isPaused
         })
         // only show pools with balances
         const hasAnyBalance = tokenPools.some((poolName) => {
+          if (!basicPools) return false
           const basicPool = basicPools?.[poolName]
-          return basicPool?.lpTokenSupply.gt(Zero) ?? false
+          if (!basicPool) return false
+          return basicPool.lpTokenSupply.gt(Zero)
```
nguyenaiden, line 148: Up to your discretion, but this is repeated more than once, so breaking out into a helper function could be beneficial here.

src/pages/Swap.tsx
```diff
@@ -604,7 +612,7 @@ function Swap(): ReactElement {
             ? "0"
             : formatUnits(
                 formState.to.value,
-                tokensMap[formState.to.symbol]?.decimals,
+                tokenSymbolToTokenMap[formState.to.symbol]?.decimals,
```
nguyenaiden, line 615: With the check on L524, the optional here could go.

ug02fast: This was one occurrence I couldn't figure out how to remove the optional check. when testing the app would break. This is one that needs to be revisited in the future.

src/pages/Swap.tsx
ug02fast: If you're talking about the console.log, David left it there so best to follow up with him.

src/pages/Swap.tsx
ug02fast: Errs w\ basicPool could be undefined.

src/pages/Swap.tsx
ug02fast: agree though I think we should revisit this in the future with a refactor to simplify some logic

nguyenaiden: Got it. I'll add all of these issues into a big `FrontEnd Clean up` epic.  Our job right now is to reduce as many extraneous optionals as possible while keeping things operational.

src/pages/Swap.tsx
nguyenaiden: @hammeiam, we will leave this in here for now, but it's poised to be removed should there not be a use for it.

src/pages/Swap.tsx
nguyenaiden: You're right. This page needs to be broken up anyway.

nguyenaiden (approved):
This is good to merge pending a very thorough manual test. Please make sure the app is operational. Good to try and run e2e locally since it's the best thing we have to verify that everything is green.

PR #1184, opened 2022-07-14
Miscellaneous POOLS_MAP references removed

ug02fast:
Miscillaneous POOLS_MAP references removed


src/pages/App.tsx
```diff
@@ -100,44 +97,30 @@ export default function App(): ReactElement {
...
-                                    exact
-                                    path={`/pools/${route}/deposit`}
-                                    render={(props) => (
-                                      <Deposit {...props} poolName={name} />
-                                    )}
-                                    key={`${route}-route-deposit`}
-                                  />
-                                ))}
-                                {pools.map(({ name }) => (
+                                {pools.map((pool) => (
```
hammeiam, line 100: A better way to do this might just be to have a dynamic route instead of an explicit one. That way instead of creating one route for each pool (and thus having to wait for all pools to be loaded, which might not work), we just have two routes for deposit and withdraw.

```
<Route exact path="/pools/:poolName/deposit" ...>
```

Then use `match.params.poolName` in the child to get pool data.

src/constants/index.ts
```diff
@@ -2199,7 +2188,7 @@ export const IS_VIRTUAL_SWAP_ACTIVE = true
 export const IS_L2_SUPPORTED = true
 export const IS_SDL_LIVE = true
 export const IS_VESDL_LIVE = true
-export const IS_POOL_REGISTRY_MIGRATION_LIVE = false
+export const IS_POOL_REGISTRY_MIGRATION_LIVE = true
```
hammeiam, line 2191: not sure if this was intentional in this PR

nkrulikovsky: To remove `POOL_MAP` from BasicProvider, don't we need to get pools from registry contract?

ug02fast: there should be no more usage off POOLS_MAP at this point.

src/pages/App.tsx
ug02fast: this is a good idea to be explored in a follow-up PR. It might work well with react-query. I will look into it and make a ticket for now.

src/hooks/useApproveAndDeposit.ts
```diff
@@ -194,7 +193,8 @@ export function useApproveAndDeposit(
         new Date().getTime() / 1000 + 60 * deadline,
       )
       if (poolName === BTC_POOL_NAME) {
-        const swapGuardedContract = effectiveSwapContract as SwapGuarded
+        const swapGuardedContract =
+          effectiveSwapContract as unknown as SwapGuarded
```
nkrulikovsky, line 197: I didn't know this kind of  type casting. 👍🏻

nkrulikovsky (approved):

src/pages/Swap.tsx
```diff
@@ -301,9 +301,9 @@ function Swap(): ReactElement {
         formStateArg.swapType === SWAP_TYPES.TOKEN_TO_SYNTH &&
         bridgeContract != null
       ) {
-        const originPool = basicPools?.[formStateArg.from.poolName]
+        const originPool = basicPools[formStateArg.from.poolName]
         if (!originPool?.metaSwapDepositAddress) {
-          error = "Unable to find metaSwap deposit address"
+          error = "Unable to find metaSwap deposit address t2s"
```
nguyenaiden, line 306: was this intentional?

nguyenaiden (approved):
Looks good to me. I tested e2e locally, and everything was green. Let's ship this, and be ready in case something comes up.

PR #1193, opened 2022-07-21
Introduce Loadables and ExpandedPools

hammeiam:
Loadables adhere to the type 
```
export interface LoadableType<T> {
  isLoading: boolean
  isLoaded: boolean
  isError: boolean
  isSuccess: boolean
  error: Error | string | null
  data?: T
}
```

ExpandedPools are just like BasicPools except they are guaranteed to have all data about their tokens and underling tokens available
```
export type ExpandedPool = Omit<
  BasicPool,
  "tokens" | "underlyingTokens" | "lpToken"
> & {
  tokens: BasicToken[]
  underlyingTokens: BasicToken[]
  lpToken: BasicToken
}
```

### DIFF
--- a/src/providers/ExpandedPoolsProvider.tsx
+++ b/src/providers/ExpandedPoolsProvider.tsx
@@ -1,0 +1,99 @@
+import { BasicPool, BasicPoolsContext } from "./BasicPoolsProvider"
+import { BasicToken, TokensContext } from "./TokensProvider"
+import {
+  EMPTY_LOADABLE,
+  LoadableType,
+  useLoadingState,
+} from "../utils/loadable"
+import React, { ReactElement, useContext } from "react"
+
+export type ExpandedPool = Omit<
+  BasicPool,
+  "tokens" | "underlyingTokens" | "lpToken"
+> & {
+  tokens: BasicToken[]
+  underlyingTokens: BasicToken[] | null
+  lpToken: BasicToken
+}
+export type ExpandedPoolsMap = Partial<Record<string, ExpandedPool>>
+export type ExpandedPools = LoadableType<ExpandedPoolsMap>
+const defaultState = {
+  ...EMPTY_LOADABLE,
+  data: {},
+}
+export const ExpandedPoolsContext =
+  React.createContext<ExpandedPools>(defaultState)
+
+export default function ExpandedPoolsProvider({
+  children,
+}: React.PropsWithChildren<unknown>): ReactElement {
+  const basicPools = useContext(BasicPoolsContext)
+  const tokens = useContext(TokensContext)
+  const { state, onStart, onSuccess } = useLoadingState<ExpandedPoolsMap>({})
+  void onStart()
+
+  const errorPools = new Set<string>()
+  let errorTokensAll = new Set<string>()
+
+  if (basicPools && tokens) {
+    const result = Object.values(basicPools || {}).reduce((acc, basicPool) => {
+      const errorTokens = new Set<string>()
+      // Tokens
+      const expandedPoolTokens = basicPool.tokens.map((tokenAddr) => {
+        const token = tokens[tokenAddr]
+        if (!token) {
+          errorTokens.add(tokenAddr)
+        }
+        return token
+      }) as BasicToken[]
+      // Underlying Tokens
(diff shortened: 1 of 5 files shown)


ug02fast (approved):

src/providers/ExpandedPoolsProvider.tsx
```diff
@@ -0,0 +1,124 @@
+import { BasicPool, BasicPoolsContext } from "./BasicPoolsProvider"
+import { BasicToken, TokensContext } from "./TokensProvider"
+import {
+  EMPTY_LOADABLE,
+  LoadableType,
+  useLoadingState,
+} from "../utils/loadable"
+import React, { ReactElement, useContext, useEffect } from "react"
+
+export type ExpandedPool = Omit<
```
nguyenaiden, line 10: For this, you're omitting the `"tokens" | "underlyingTokens" | "lpToken"` keys and readding it after. Could we potentially just overwrite here?

src/providers/ExpandedPoolsProvider.tsx
```diff
@@ -0,0 +1,124 @@
...
+    },
+  )
+
+  useEffect(() => {
+    onStart()
+  }, [onStart])
+
+  useEffect(() => {
+    const errorPools = new Set<string>()
+    let errorTokensAll = new Set<string>()
```
nguyenaiden, line 48: Up to your discretion. Not a fan of reassignments, so I think we can do something like this:
```useEffect(() => {
    const errorPools: string[] = []
    const errorTokensAll: string[] = []
    if (basicPools && tokens) {
      const result = Object.values(basicPools || {}).reduce(
        (acc, basicPool) => {
          const errorTokens: string[] = []
          // Tokens
          const expandedPoolTokens = basicPool.tokens.map((tokenAddr) => {
            const token = tokens[tokenAddr]
            if (!token) {
              errorTokens.push(tokenAddr)
            }
            return token
          }) as BasicToken[]
          // Underlying Tokens
          const expandedUnderlyingPoolTokens = basicPool.isMetaSwap
            ? (basicPool.underlyingTokens.map((tokenAddr) => {
                const token = tokens[tokenAddr]
                if (!token) {
                  errorTokens.push(tokenAddr)
                }
                return token
              }) as BasicToken[])
            : null
          // LP Token
          const expandedLpToken = tokens[basicPool.lpToken]
          if (!expandedLpToken) {
            errorTokens.push(basicPool.lpToken)
          }

          // If any tokens aren't found, invalidate the whole pool
          if (errorTokens.length > 0) {
            errorTokensAll.concat(errorTokens)
            errorPools.push(basicPool.poolName)
            return acc
          }
          const expandedPool = {
            ...basicPool,
            tokens: expandedPoolTokens,
            underlyingTokens: expandedUnderlyingPoolTokens,
            lpToken: expandedLpToken,
          } as ExpandedPool
          return {
            byName: {
              ...acc.byName,
              [basicPool.poolName]: expandedPool,
            },
            byAddress: {
              ...acc.byAddress,
              [basicPool.poolAddress]: expandedPool,
            },
          }
        },
        { byName: {}, byAddress: {} },
      )

      if (errorTokensAll.length > 0) {
        const tokensString = [...new Set(errorTokensAll.values())].join(", ")
        const poolsString = [...new Set(errorPools.values())].join(", ")
        console.error(
          `Error loading tokens [${tokensString}] in pools [${poolsString}]`,
        )
      }

      onSuccess(result)
    }
  }, [basicPools, onSuccess, tokens])

nguyenaiden (approved):
2 comments, but LGTM!

src/providers/ExpandedPoolsProvider.tsx
hammeiam: I'm not sure! Makes sense that we could though. I'll give it a shot.

src/providers/ExpandedPoolsProvider.tsx
hammeiam: You're right, I like this better. Get those `let`s outta here.

src/providers/ExpandedPoolsProvider.tsx
hammeiam: nope didn't work

PR #1195, opened 2022-07-23
react-query initial bring in with respective upgrades 

ug02fast:
pre-reqs for bringing in `react-query` are:

upgrade to latest `react-scripts`
upgrade react + react-dom versions
fix few types
change initial render to adhere to new React 18 approach
webpack hack on `metamaskIcon` import due to svg loading changes.

image


src/components/Web3Status.tsx
```diff
@@ -41,9 +41,13 @@ const Web3Status = (): ReactElement => {
         endIcon={account && <Identicon />}
       >
         <Typography variant="body1" whiteSpace="nowrap">
-          {account
-            ? udName || ensName || shortenAddress(account)
-            : t("connectWallet")}
+          {isLoading ? (
+            <Typography>Loading...</Typography>
```
nkrulikovsky, line 45: Seems like don't need `typography` because it is already wrapped with `typography`.

src/hooks/usePoolData.ts
```diff
@@ -16,6 +16,7 @@ import { UserStateContext } from "./../providers/UserStateProvider"
 import { Zero } from "@ethersproject/constants"
 import { parseUnits } from "@ethersproject/units"
 import { useActiveWeb3React } from "."
+// import { useBasicTokens } from "../providers/TokensProvider"
```
nkrulikovsky, line 19: Is this comment intentional?

src/components/Web3Status.tsx
ug02fast: Ok, I forgot to remove this. We should get some sort of design for loading states.

src/hooks/usePoolData.ts
ug02fast: Going to remove, this was before the ExpandedPool Provider so is a little outdated.

src/components/TopMenu.tsx
```diff
@@ -178,7 +178,9 @@ function MenuList() {
   const { t } = useTranslation()
   const { chainId } = useActiveWeb3React()
   const gaugesAreActive = areGaugesActive(chainId)
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const { pathname } = useLocation()
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
ug02fast, line 183: Lost types or incompatible w\ latest version. Will have to prioritize router package upgrade.

src/components/PoolOverview.tsx
```diff
@@ -58,6 +58,7 @@ export default function PoolOverview({
     sdlPerDay: formatBNToShortString(poolData?.sdlPerDay || Zero, 18),
   }
   const hasShare = !!userShareData?.usdBalance.gt(Zero)
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
ug02fast, line 61: Lost types or incompatible w\ latest version. Will have to prioritize router package upgrade.

src/hooks/useApproveAndDeposit.ts
```diff
@@ -57,7 +57,7 @@ export function useApproveAndDeposit(
     infiniteApproval,
   } = useSelector((state: AppState) => state.user)
   const basicPools = useContext(BasicPoolsContext)
-  const tokens = useContext(TokensContext)
+  const { data: tokens } = useBasicTokens()
```
ug02fast, line 60: small example of tokens coming from query instead of Context

src/hooks/useBasicTokens.ts
```diff
@@ -0,0 +1,172 @@
...
+      targetTokenAddresses.add(
+        VOTING_ESCROW_CONTRACT_ADDRESS[chainId].toLowerCase(),
+      )
+    }
+    const tokenInfos = await getTokenInfos(
+      ethCallProvider,
+      chainId,
+      Array.from(targetTokenAddresses),
+    )
+    if (!tokenInfos) return
```
hammeiam, line 101: Should we add a task to rewrite these fetcher functions to return errors if they're unsuccessful? Currently we return null in a lot of places, but the loadable / useQuery pattern gives us the ability to better handle and surface errors

src/hooks/useBasicTokens.ts
```diff
@@ -0,0 +1,172 @@
...
+  const lpTokens = new Set()
+  const tokenType: Partial<{ [tokenAddress: string]: PoolTypes }> = {}
+
+  return useQuery(["tokens"], async () => {
+    if (!chainId || !basicPools || !ethCallProvider) {
+      return null
+    }
+    const targetTokenAddresses = new Set(
+      Object.values(basicPools)
+        .map((pool) => {
```
hammeiam, line 53: I feel I was probably the one who wrote the code originally, but I'd prefer not to have side effects in map or reduce blocks - forEach is fine.

src/hooks/useBasicTokens.ts
```diff
@@ -0,0 +1,172 @@
...
+    const [nameResults, symbolResults, decimalsResults] = await Promise.all([
+      chunkedTryAll(nameCalls, ethCallProvider, 30),
+      chunkedTryAll(symbolCalls, ethCallProvider, 30),
+      chunkedTryAll(decimalsCalls, ethCallProvider, 30),
+    ])
+    const results = lowercaseTokenAddresses.reduce((acc, address, index) => {
+      const name = nameResults[index]
+      const symbol = symbolResults[index]
+      const decimals = decimalsResults[index] // could be 0
+      if (name && symbol && decimals != null) {
```
hammeiam, line 149: The "else" of this "if" is where the error would be created

src/hooks/useENS.ts
```diff
@@ -1,44 +1,38 @@
...
 import { ChainId } from "../constants"
 import { ethers } from "ethers"
 import { useActiveWeb3React } from "."
 
-type ReturnType = { ensName: string | null }
+type ENS = string | null
 
-export const useENS = (address: string | null | undefined): ReturnType => {
-  const [ensName, setENSName] = useState<string | null>(null)
+export const useENS = (address: string): UseQueryResult<ENS> => {
```
hammeiam, line 9: Better yet, I'm pretty sure wagmi / rainbow has a built in useENS hook

src/hooks/useMulticallProvider.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { MulticallProvider } from "../types/ethcall"
+import { getMulticallProvider } from "../utils"
+import { useActiveWeb3React } from "../hooks"
+
+export const useMulticallProvider = (): UseQueryResult<MulticallProvider> => {
+  const { chainId, library } = useActiveWeb3React()
+
+  return useQuery(["multicallProvider"], async () => {
+    if (!library || !chainId) throw new Error("library or chainId not found")
```
hammeiam, line 11: UseQuery just let's you throw and then handles it? I love that

src/hooks/usePoolData.ts
```diff
@@ -132,6 +132,7 @@ export default function usePoolData(name?: string): PoolDataHookReturnType {
     async function getSwapData(): Promise<void> {
       const expandedPool = poolName ? expandedPools.byName[poolName] : null
       if (
+        // isLoading ||
```
hammeiam, line 135: Rm?

src/pages/App.tsx
```diff
@@ -80,97 +90,108 @@ export default function App(): ReactElement {
...
-                        <RewardsBalancesProvider>
-                          <LocalizationProvider dateAdapter={AdapterDateFns}>
-                            <AppContainer>
-                              <TopMenu />
-                              <Switch>
-                                <Route exact path="/" component={Swap} />
-                                <Route exact path="/pools" component={Pools} />
-                                {pools.map((pool) => (
+    <QueryClientProvider client={queryClient}>
+      <ReactQueryDevtools initialIsOpen />
```
hammeiam, line 94: Can we include this only in dev environments?

src/pages/App.tsx
```diff
@@ -80,97 +90,108 @@ export default function App(): ReactElement {
...
-                                        poolName={pool.poolName}
-                                      />
-                                    )}
-                                    key={`${pool.poolName}-deposit`}
+                                    path="/pools"
+                                    component={Pools}
                                   />
-                                ))}
-                                {pools.map((pool) => (
+                                  {pools.map((pool) => (
```
hammeiam, line 116: Rather than mapping over pools, can we just pass the poolName route param to the deposit or withdraw component? That way we'll have 2 routes instead of (n)

hammeiam (approved):
Exciting!

src/hooks/useMulticallProvider.ts
ug02fast: Yes, when within the useQuery asynchronous fn parameter. It will then manage the error.

src/pages/App.tsx
ug02fast: It automatically does that.

src/hooks/useBasicTokens.ts
ug02fast: Yeah these queries can be improved. This is a bare minimum port example.

src/hooks/useBasicTokens.ts
ug02fast: There are about 3 or more diff ways we can handle failed calls and what logic is inside the useQuery vs what's outside of it. The Promise.all above might throw the error in this case though not too sure. Let's revisit this in a following PR to iron down a good way to construct the queries.

src/pages/App.tsx
ug02fast: big

PR #1196, opened 2022-07-24
TS Upgrade + corresponding changes for new rules.

ug02fast:
This brings in the TS upgrade and necessary eslint/tslint changes.

closes: https://github.com/saddle-finance/saddle-frontend/pull/1163


src/pages/CreatePool/index.tsx
```diff
@@ -201,6 +201,41 @@ export default function CreatePool(): React.ReactElement {
     tokenInfo,
   ])
 
+  const onTokenInputBlur = async (
```
ug02fast, line 204: just moved fn logic out of return statement here

hammeiam (approved):
You're doing gods work 🙏

PR #1204, opened 2022-07-29
Minichef pids updated for Opt Frax pools.

ug02fast:
includes arb pid fix


penandlim (commented):
Need to add Minichef address on optimism network at `0x220d6bEedeA6a6317DaE19d39cd62EB7bb0ae5e4`

But looks good otherwise

hammeiam:
Nice fix! I forget how minichef works, but I hope we could enumerate through the on chain pids like we do with gauges so we can remove these hard coded values, too.

PR #1219, opened 2022-08-03
Date picker patch

nguyenaiden:
***Description***: Upon the `date-fns` package update from `2.25` -> `2.29`,  `intervalToDuration` method will throw a `RangeError` if the end date is before the start date. This PR addresses that issue by switching the start and end key of the input `Interval` object of that method.


ug02fast (approved):

nguyenaiden:
Closing as the alternative patch was merged instead.

PR #1229, opened 2022-08-05
Disable ramp A value and end data when date value is zero

nkrulikovsky:
## Description

For pool that have never changed their A parameter, ramping end date value is zero(Big number). In this case, we don't need to display this value

image

image


hammeiam (approved):
Will this also address the future A value being incorrectly shown as 10,000? Less importantly, do we want to display futureATimes from the past?

nguyenaiden (commented):
Could you add a screen capture in the PR description?

hammeiam (approved):
Thanks for the precision update!

nkrulikovsky:
@nguyenaiden 
Added screenshot

PR #1231, opened 2022-08-07
Calculate SDL earned per day, log in dev

hammeiam:
Take the SDL rate of all gauges times the user share per gauge to get the user's SDL earned per day. Log to the console in dev. We can show this to the user once we have a design for it.


src/pages/Farm/Farm.tsx
```diff
@@ -47,54 +125,8 @@ export default function Farm(): JSX.Element {
         <FarmListHeader />
       </Box>
 
-      {Object.values(gauges)
```
hammeiam, line 50: All this just got moved to above where UI is returned, for cleanliness

ug02fast (approved):

PR #1233, opened 2022-08-09
On-chain vote

nkrulikovsky:
## Description

- On-chain vote section under the Gauge graph
- My votes lists
- User can reset vote after 10 days later after vote

## Screenshot

https://user-images.githubusercontent.com/75422800/185504970-53c232c4-6de9-4f78-a43d-f65fb493b204.mov

## How to test this feature?

- Fork mainnet 
  On saddle-contract repo, run `npx hardhat node --fork mainnet --fork-ignore-unknown-tx-type true`

- After fork, run `npx hardhat run scripts/forked_mainnet/enableOnChainGaugeVoting.ts --network localhost --include-network mainnet`
-


public/locales/en.json
```diff
@@ -315,15 +322,18 @@
...
   "virtualPrice": "Virtual price",
   "virtualSwap": "Virtual Swap",
   "virtualSwapSynthToSynthInfo": "This is a Synth-to-Synth Virtual Swap. You will see your swapped tokens in your wallet immediately after the swap executes, but must wait for up to 6 minutes before being able to transfer them.",
   "volume": "Volume",
-  "vote": "Vote",
+  "vote": "Vote completed",
   "voteEscrowSDL": "Vote escrow SDL",
+  "voteForNextWeek": "Vote for next week",
   "votePeriod": "Vote period",
+  "voteThisGauge": "Vote this gauge",
```
ug02fast, line 333:
```suggestion
  "voteThisGauge": "Vote for this gauge",
```
? @hammeiam

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const gaugeControllerContract = useGaugeControllerContract(true)
+  const [voteWeightInputError, setVoteWeightInputError] = useState<{
+    hasError: boolean
+    errorText: string
+  }>({ hasError: false, errorText: " " })
+
+  const gaugeNames = useMemo(
+    () =>
+      Object.keys(gauges)
+        .filter((lpAddress) => !gauges[lpAddress]?.isKilled)
```
ug02fast, line 63: can you early null check/return `gauges[lpAddress].<property>` so the value can't be `undefined` during this execution?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+        .map((lpAddress) => {
+          const gaugeName = gauges[lpAddress]?.gaugeName
```
ug02fast, line 65: can you early null check/return gauges[lpAddress].<property> so the value can't be undefined during this execution?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+          const gaugeAddress = gauges[lpAddress]?.address
```
ug02fast, line 66: can you early null check/return gauges[lpAddress].<property> so the value can't be undefined during this execution?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+        }),
+    [gauges],
+  )
+
+  const getVoteUsed = useCallback(async () => {
+    if (account) {
+      const voteUsed = await gaugeControllerContract?.[
+        "vote_user_power(address)"
+      ](account)
+      setVoteUsed(voteUsed?.toNumber())
```
ug02fast, line 81: can you early null check/return these?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+  }, [account, gaugeControllerContract])
+
+  useEffect(() => {
+    if (account && gaugeControllerContract) {
+      gaugeControllerContract.on(
+        "VoteForGauge",
+        (voteDate, fromAddress, gaugeAddress, voteWeight) => {
+          const gaugeName = gaugeNames.find(
+            (gaugeName) =>
+              gaugeName?.address.toLowerCase() ===
```
ug02fast, line 92: early null check

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+      const voteUsed = await gaugeControllerContract?.[
+        "vote_user_power(address)"
+      ](account)
+      setVoteUsed(voteUsed?.toNumber())
+    }
+  }, [account, gaugeControllerContract])
+
+  useEffect(() => {
+    if (account && gaugeControllerContract) {
+      gaugeControllerContract.on(
```
ug02fast, line 87: cool stuff

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+      )
+    }
+    return () => {
+      gaugeControllerContract?.removeAllListeners()
+    }
+  }, [gaugeControllerContract, gaugeNames, votes, getVoteUsed, account])
+
+  const handleVote = async () => {
+    const voteWeightPercent = parseFloat(voteWeightToSubmit)
+    if (!Number.isNaN(voteWeightPercent)) {
```
ug02fast, line 120: Why would this not be a number?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+          setVoteWeightToSubmit("") // Initialize vote weight input
+        }
+      } catch (error) {
+        console.log("error on vote ==>", error)
+      }
+    }
+  }
+
+  return (
+    <div>
```
ug02fast, line 140: can you use react.fragment instead of div here?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+      <Stack spacing={2} px={2} pt={2} mb={2}>
+        <Typography variant="h2" textAlign="center">
+          {t("voteForNextWeek")}
+        </Typography>
+        <Box display="flex" justifyContent="space-between">
+          <Box display="flex">
+            <Typography component="label">
+              {t("yourVotingPower")}:{" "}
```
ug02fast, line 148: add padding or margin to Typograph/Box instead of putting manual " " space there.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+                {...params}
+                placeholder={t("chooseGauge")}
+              />
+            )}
+            onChange={(event, newGaugeName) => {
+              if (newGaugeName) {
+                setSelectedGauge(newGaugeName)
+              }
+            }}
+            //value of selected gauge
```
ug02fast, line 193: prop name and value good enough comment for me personally.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+                    parseFloat(event.target.value) + (voteUsed ?? 0) / 100
+                  if (estimatedWeight > 100) {
+                    setVoteWeightInputError({
+                      hasError: true,
+                      errorText: "Used too much power",
+                    })
+                  } else {
+                    setVoteWeightInputError({
+                      hasError: false,
+                      errorText: " ",
```
ug02fast, line 218: Can you add a reasonable error message here?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,254 @@
...
+    if (!Number.isNaN(voteWeightPercent)) {
+      try {
+        if (veSdlBalance.isZero()) {
+          setAlertMessage("You need veSDL to vote")
+        } else if (selectedGauge?.address && gaugeControllerContract) {
+          const txn = await gaugeControllerContract.vote_for_gauge_weights(
+            selectedGauge.address,
+            Math.round(voteWeightPercent * 100), //convert percent to bps
+          )
+          void enqueuePromiseToast(1, txn.wait(), "vote")
```
ug02fast, line 129: does toast work when using `void` instead of `await`?

ug02fast (changes requested):
great job! mostly curious if you can early return the falsy values until data loaded in instead of passing potentially `undefined/null` values after the ?.

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: We should set `space` here because we don't need to display message if estimated weight is less than `100%`

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: You are correct. We should use `await`

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Good callout!
Replace this logic with `reduce` to avoid `undefined`

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Fixed

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: This checks `Nan` value of `voteWeightPercent`

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Fixed

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: fixed

src/pages/VeSDL/OnChainVote.tsx
ug02fast: a percent is always a number so it should always be between 0-1 which is why checking if it's a number is weird to me from reading standpoint

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -61,24 +62,25 @@ export default function OnChainVote({
     () =>
       Object.keys(gauges)
         .filter((lpAddress) => !gauges[lpAddress]?.isKilled)
-        .map((lpAddress) => {
-          const gaugeName = gauges[lpAddress]?.gaugeName
-          const gaugeAddress = gauges[lpAddress]?.address
+        .reduce((acc, curr) => {
```
ug02fast, line 65: instead of changing this to reduce, can you check 
```
if (!gauges[lpAddress]) return ____ (whatever the appropriate "empty" value is)
```
reduce isn't very functional

ug02fast (changes requested):

src/pages/VeSDL/VoteHistory.tsx
```diff
@@ -0,0 +1,163 @@
...
+  gaugeNames,
+}: Props) {
+  const theme = useTheme()
+  const { account } = useActiveWeb3React()
+  const [voteUsed, setVoteUsed] = useState<number | undefined>()
+  const [voteLists, setVoteLists] = useState<VoteList[]>([])
+
+  const getVoteUsed = useCallback(async () => {
+    if (account) {
+      const voteUsed = await gaugeControllerContract?.[
```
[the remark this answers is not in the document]
nkrulikovsky: I have updated code to check gaueControllerContract before loading `On-Chain vote` component.

src/pages/VeSDL/VoteHistory.tsx
```diff
@@ -0,0 +1,163 @@
...
+  gaugeNames: (GaugeName | undefined)[]
+}
+
+export default function VoteHistory({
+  gaugeControllerContract,
+  gaugeNames,
+}: Props) {
+  const theme = useTheme()
+  const { account } = useActiveWeb3React()
+  const [voteUsed, setVoteUsed] = useState<number | undefined>()
```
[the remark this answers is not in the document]
nkrulikovsky: You are right.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,160 @@
...
+} from "@mui/material"
+import React, { useContext, useMemo, useState } from "react"
+import ArrowDownIcon from "@mui/icons-material/KeyboardArrowDown"
+import { GaugeContext } from "../../providers/GaugeProvider"
+import VoteHistory from "./VoteHistory"
+import { enqueuePromiseToast } from "../../components/Toastify"
+import { useGaugeControllerContract } from "../../hooks/useContract"
+
+interface OnChainVoteProps {
+  veSdlBalance: string
```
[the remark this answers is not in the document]
nkrulikovsky: `OnChainVote`is receiving  this value after converting `BigNumber` to `string` because we only display this value on `OnChainvote`

src/pages/VeSDL/VoteHistory.tsx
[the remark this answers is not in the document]
nkrulikovsky: If we set vote used as `0`, then it will show `0` while fetching data from contract. It is good to remove `undefined` from value but this value is not sued on other function, so I think, it is rather to hide this value when there is no vote used value.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const theme = useTheme()
+  const { gauges } = useContext(GaugeContext)
+  const [selectedGauge, setSelectedGauge] = useState<GaugeName | null>(null)
+  const [voteWeightToSubmit, setVoteWeightToSubmit] = useState<string>("")
+  const gaugeControllerContract = useGaugeControllerContract(true)
+
+  const gaugeNames = useMemo(
+    () =>
+      Object.keys(gauges)
+        .filter((lpAddress) => !gauges[lpAddress]?.isKilled)
```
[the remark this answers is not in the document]
nkrulikovsky: reason is because Gauges type is `Partial<{[lpTokenAddress: string]: Gauge}>`

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: `voteWeightToSubmit` is text we input, `parseFloat('abcedf')` returns `Nan`, so it is worth to check this value whether it is `Nan` or not.
Also, FYI, the percent number is in 0~100% in here.

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Updated from reduce to using foreach

src/pages/VeSDL/OnChainVote.tsx
ug02fast: it will be better to do `number` attribute on that input or use regex to only allow numbers

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -58,22 +63,21 @@ export default function OnChainVote({
...
-            acc?.push({
-              address: gaugeAddress,
-              gaugeName: gaugeName,
-            })
-          return acc
-        }, [] as GaugeName[]),
-    [gauges],
-  )
+  const gaugeNames = useMemo(() => {
+    const gaugeNameObj = {} as GaugeName
```
ug02fast, line 67: instead of pushing values into an object can you just return the final result as a map? also what is the point of useMemo here?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,286 @@
...
+        if (gaugeName && gaugeAddress)
+          gaugeNameObj[gaugeAddress] = {
+            address: gaugeAddress,
+            gaugeName: gaugeName,
+          }
+      })
+    return gaugeNameObj
+  }, [gauges])
+
+  const getVoteUsed = useCallback(async () => {
```
ug02fast, line 82: instead of leveraging useCallback + useEffect can you use react-query?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,286 @@
...
+    if (account) {
+      const voteUsed = await gaugeControllerContract[
+        "vote_user_power(address)"
+      ](account)
+      setVoteUsed(voteUsed.toNumber())
+    }
+  }, [account, gaugeControllerContract])
+
+  const getFilteredVotes = useCallback(async () => {
```
ug02fast, line 91: instead of leveraging useCallback + useEffect can you use react-query?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,286 @@
...
+          }
+        }
+        return acc
+      },
+      {} as VotesType,
+    )
+    setVotes(filteredVotes)
+  }, [account, gaugeControllerContract, gaugeNames])
+
+  useEffect(() => {
```
ug02fast, line 116: instead of leveraging useCallback + useEffect can you use react-query?

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Updated with react-query.

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Updated.

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: updated

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: I made `gaugeName` to select `gauge Name` buy its address easily. We need to display vote status for each Gauge rather than vote history.
From event of gauge vote, we can get `user`, `time`, weight`, `gauge_addr`.

So, it would be good to make gaugeName as object rather than map to find the gauge name easily.
And I used `useMemo` to prevent the recalculate the `gaugeName` every time when state of onChain vote change.
We only need to calculate when `gauges` change.

ug02fast (commented):
a few considerations on top of the `gaugeName` discussion

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+          }
+      })
+    return gaugeNameObj
+  }, [gauges])
+
+  const getVoteUsed = async () => {
+    if (!account) return
+    return gaugeControllerContract["vote_user_power(address)"](account)
+  }
+  const getFilteredVotes = async () => {
```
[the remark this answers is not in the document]
nkrulikovsky: query function is having type of `queryFn: (context: QueryFunctionContext) => Promise<TData>`.
`getVoteUsed` is having `const getVoteUsed: () => Promise<BigNumber | undefined>` , so I believe we don't need to use `await`

ug02fast (approved):

src/hooks/useContract.ts
```diff
@@ -334,15 +334,17 @@ export function useLPTokenContract(
...
+): GaugeController | null {
   const { chainId } = useActiveWeb3React()
   const contractAddress = chainId
     ? GAUGE_CONTROLLER_ADDRESSES[chainId]
     : undefined
   return useContract(
     contractAddress,
     GAUGE_CONTROLLER_ABI,
-    false,
+    withSigner ? true : false,
```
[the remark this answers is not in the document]
nkrulikovsky: @ug02fast 
You are right. I changed this logic a little to remove the withSigner parameter on `useGaugeControllerContract`

src/hooks/useContract.ts
```diff
@@ -335,14 +335,14 @@ export function useLPTokenContract(
...
-  const { chainId } = useActiveWeb3React()
+  const { account, chainId } = useActiveWeb3React()
   const contractAddress = chainId
     ? GAUGE_CONTROLLER_ADDRESSES[chainId]
     : undefined
   return useContract(
     contractAddress,
     GAUGE_CONTROLLER_ABI,
-    false,
+    account ? true : false,
```
ug02fast, line 345: Is account necessary here?

nkrulikovsky: Good question. I think we can give `undefined` for this parameter.
But I am not exactly sure why we used `false` instead of `undefined`

ug02fast: Does anything change if an account is not included in this?

hammeiam:
The UI looks great and works well! 

The contract prevents users from voting for the same gauge more than once in 10 days - could you either display a warning to the user before they vote, and also display the error message that the contract reverts with if the user insists on trying to vote again?

src/pages/VeSDL/GaugeVote.tsx
```diff
@@ -1,85 +1,40 @@
...
-            })}
-          </TableBody>
-        </Table>
-      </TableContainer>
-      <Typography textAlign="end" mt={1}>
-        <Link color="inherit" href={generateSnapshotVoteLink()} target="_blank">
-          {t("viewAllVote")}
-        </Link>
-      </Typography>
+      {!isGaugesLoading && gaugeControllerContract?.signer ? (
```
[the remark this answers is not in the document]
nkrulikovsky: Now, we are having `withSignerIfPossible` to give `account` to `getContract`.
We can get account from 'useActiveWeb3React`, it means we don't need ` withSignerIfPossible` parameter on `useContract` to check.
On the contract returning from useContract, we can check whether given contract is containing `signer` or not.

Also, useContract is using on  to create specific contract, so removing this parameter will make more clean.

src/hooks/useContract.ts
nkrulikovsky: Now, we are having `withSignerIfPossible` to give `account` to `getContract`.
We can get account from `useActiveWeb3React`, it means we don't need ` withSignerIfPossible` parameter on `useContract` to check.
On the contract returning from useContract, we can check whether given contract is containing `signer` or not.

Also, useContract is using on  to create specific contract, so removing this parameter will make more clean.

nkrulikovsky:
> The contract prevents users from voting for the same gauge more than once in 10 days - could you either display a warning to the user before they vote, and also display the error message that the contract reverts with if the user insists on trying to vote again?

Now, after user vote on certain `gauge` the vote button

src/pages/VeSDL/GaugeVote.tsx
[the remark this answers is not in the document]
nkrulikovsky: With this we can remove `withSignerIfPossible`. Instead, we can return error message when there is no account if we need account for that contract.

src/pages/VeSDL/GaugeVote.tsx
```diff
@@ -1,85 +1,40 @@
...
-
+interface GaugeVoteProps {
+  veSdlBalance: BigNumber
+}
+export default function GaugeVote({
+  veSdlBalance,
+}: GaugeVoteProps): JSX.Element {
   const { t } = useTranslation()
+  const { gauges } = useContext(GaugeContext)
+  const isGaugesLoading = Object.keys(gauges).length === 0
```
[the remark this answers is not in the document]
nkrulikovsky: I am updating this logic which implementing the loading skeleton on another branch.

src/pages/VeSDL/Votes.tsx
```diff
@@ -0,0 +1,119 @@
...
+    try {
+      if (gaugeControllerContract) {
+        const txn = await gaugeControllerContract.vote_for_gauge_weights(
+          gaugeAddress,
+          0, // to reset vote with zero
+        )
+        void enqueuePromiseToast(1, txn.wait(), "vote")
+      }
+    } catch (error) {
+      console.error("error on vote ==>", error)
```
hammeiam, line 59: You should also notify the user via toast that we weren't able to complete the action

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+          />
+          <Box display="flex" gap={4} my={2}>
+            <TextField
+              variant="standard"
+              label={t("votingWeight")}
+              type="text"
+              value={voteWeightToSubmit}
+              onChange={(event) => {
+                if (isNumberOrEmpty(event.target.value)) {
+                  const estimatedWeight =
```
hammeiam, line 200: Does this work? I'd expect JS to use the standard order of operations in math: parenthesis, multiplication, division, addition, subtraction. eg

`99 + 99 / 100 = 99.99`

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  const handleVote = async () => {
+    if (voteWeightToSubmit && isNumberOrEmpty(voteWeightToSubmit) && chainId) {
+      try {
+        if (veSdlBalance.isZero()) {
+          setAlertMessage("You need veSDL to vote")
+        } else if (selectedGauge?.address && gaugeControllerContract) {
+          const txn = await gaugeControllerContract.vote_for_gauge_weights(
+            selectedGauge.address,
+            Math.round(parseFloat(voteWeightToSubmit) * 100), //convert percent to bps
```
hammeiam, line 120: Using `Math.round` may lead to the txn not being able to execute as expected. If the user has `n`% voting power remaining, and we round their input up to `n + 1`, then they would appear to be voting with too much power.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+          }
+      })
+    return gaugeNameObj
+  }, [gauges])
+
+  const getVoteUsed = () => {
+    if (!account) return
+    return gaugeControllerContract["vote_user_power(address)"](account)
+  }
+  const getFilteredVotes = () => {
```
hammeiam, line 85: Should these be wrapped in a `useCallback`?

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+  [gaugeAddress: string]: {
+    gaugeName: string
+    weight: BigNumber
+    voteDate: number
+  }
+}
+
+export default function OnChainVote({
+  veSdlBalance,
+  gauges,
```
hammeiam, line 48: Is there a specific reason to pass gauges as a prop? It's already accessible through context.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+import { GaugeController } from "../../../types/ethers-contracts/GaugeController"
+import VoteHistory from "./Votes"
+import { enqueuePromiseToast } from "../../components/Toastify"
+import { useActiveWeb3React } from "../../hooks"
+import { useQuery } from "@tanstack/react-query"
+import { useTranslation } from "react-i18next"
+
+interface OnChainVoteProps {
+  veSdlBalance: BigNumber
+  gauges: Partial<{
```
hammeiam, line 26: this type is the same as `LPTokenAddressToGauge`

src/hooks/useContract.ts
```diff
@@ -342,7 +342,7 @@ export function useGaugeControllerContract(): GaugeController | null {
   return useContract(
     contractAddress,
     GAUGE_CONTROLLER_ABI,
-    false,
+    undefined,
```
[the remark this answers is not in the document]
nkrulikovsky: We should give  `undefined` because `useGaugeControllerContract` needs signer when user vote.

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: I believe `query function` don't need to be wrapped with `useCallback`.

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Because we still don't have loading status for `gauges` on context, I am checking gauge load status by following on parent component.
```
const isGaugesLoading = Object.keys(gauges).length === 0
```

Will update this logic while update component with `useQuery`

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+    const voteFilter = gaugeControllerContract.filters.VoteForGauge(
+      null,
+      null,
+      null,
+      null,
+    )
+    return gaugeControllerContract.queryFilter(voteFilter)
+  }
+
+  const { data: voteUsed } = useQuery(["voteUsed"], getVoteUsed)
```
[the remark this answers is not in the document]
nkrulikovsky:
```
   const getVoteUsed = () => {
    if (!account) return
    return gaugeControllerContract["vote_user_power(address)"](account)
  }
```
Above `getVoteUsed` is a promise call. To get `voteUsed` from gaugeControllerContract, originally, we need to use useEffet to fetch value using await, but with the help of useQuery, we don't need to use useEffect and useCallback.

src/pages/VeSDL/Votes.tsx
nkrulikovsky: ✅

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+              />
+            )}
+            onChange={(event, newGaugeName) => {
+              if (newGaugeName) {
+                setSelectedGauge(newGaugeName)
+              }
+            }}
+            value={selectedGauge}
+            // check whether the current value belong to options
+            isOptionEqualToValue={(option, value) =>
```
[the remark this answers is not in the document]
nkrulikovsky: Fixed.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,247 @@
...
+          const txn = await gaugeControllerContract.vote_for_gauge_weights(
+            selectedGauge.address,
+            Math.round(parseFloat(voteWeightToSubmit) * 100), //convert percent to bps
+          )
+          await enqueuePromiseToast(chainId, txn.wait(), "vote")
+          setSelectedGauge(null) // Initialize selectedGauge
+          setVoteWeightToSubmit("") // Initialize vote weight input
+        }
+      } catch (error) {
+        console.error("error on vote ==>", error)
```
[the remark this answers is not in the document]
nkrulikovsky: Updated

src/pages/VeSDL/Votes.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    </Button>
+                  </TableCell>
+                </VoteTableRow>
+              )
+            })}
+          </TableBody>
+        </Table>
+        {!!voteUsed && (
+          <Typography mt={1}>
+            {t("totalWeightUsed")}: {voteUsed.toNumber() / 100}%
```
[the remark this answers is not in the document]
nkrulikovsky: Updated

src/pages/VeSDL/OnChainVote.tsx
nkrulikovsky: Fixed.

src/pages/VeSDL/Votes.tsx
```diff
@@ -0,0 +1,119 @@
...
+          </TableHead>
+          <TableBody>
+            {Object.keys(votes).map((gaugeAddress) => {
+              const vote = votes[gaugeAddress]
+              const passed10Days =
+                Date.now() / 1000 > vote.voteDate + tenDaysInSecond
+              return (
+                <VoteTableRow key={gaugeAddress}>
+                  <TableCell>{vote.gaugeName}</TableCell>
+                  <TableCell>{vote.weight.toNumber() / 100}%</TableCell>
```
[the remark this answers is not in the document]
nkrulikovsky: updated.

src/constants/index.ts
```diff
@@ -2271,6 +2271,7 @@ export const IS_VIRTUAL_SWAP_ACTIVE = true
 export const IS_L2_SUPPORTED = true
 export const IS_SDL_LIVE = true
 export const IS_VESDL_LIVE = true
+export const IS_ON_CHAIN_VOTE_LIVE = true //TODO: shoudl be false
```
hammeiam, line 2274: remember to turn this off before landing :)

hammeiam (approved):
lgtm! @nguyenaiden could you take a look through the lens of x-chain gauges? We'll be prioritizing landing this PR first, since we'll still need to wait for an audit of the x-chain gauge contracts.

src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -0,0 +1,251 @@
...
+        )}
+        <Box
+          border={`1px solid ${theme.palette.other.divider}`}
+          px={2}
+          pt={2}
+          borderRadius="10px"
+        >
+          <Autocomplete
+            id="gauge-names"
+            options={Object.keys(gaugeNames).map((key) => gaugeNames[key])}
```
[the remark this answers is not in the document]
nkrulikovsky: 🧠

hammeiam:
Nit - we should disable voting for users with 0 veSDL balance 

Screen Shot 2022-11-04 at 7 09 58 PM

PR #1234, opened 2022-08-10
Use graphql to filter snapshot results

hammeiam:


src/utils/getSnapshotVoteData.ts
```diff
@@ -20,10 +21,11 @@ export default function getSnapshotVoteData(
    */
   const query = `{
     proposals (
-      first: 20,
+      first: 10,
       skip: 0,
       where: {
         space_in: ["saddlefinance.eth"],
+        title_contains: "Gauge reward allocation"
```
nguyenaiden, line 28: Can we make sure that this is case-insensitive?

nguyenaiden (approved):
+1. Just need to verify case sensitivity.

PR #1238, opened 2022-08-12
Use the correct withdraw method based on current and unlock date timestamps

nguyenaiden:
***Description:*** Currently, the app is using `voteEscrowContract.force_withdraw()` for all veSDL unlocking operations. This method will throw a "lock expired" error if the current timestamp is after the unlock date.

Note: Since we're not using the block timestamp, there is a 10-15 second window where the withdraw() method will throw a similar error.

### DIFF
--- a/src/pages/VeSDL/index.tsx
+++ b/src/pages/VeSDL/index.tsx
@@ -12,6 +12,7 @@ import {
   Tooltip,
   Typography,
 } from "@mui/material"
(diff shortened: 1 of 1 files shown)


hammeiam (approved):
lgtm

nkrulikovsky (approved):

ug02fast (approved):

PR #1239, opened 2022-08-16
usePoolRegistry wagmi

ug02fast:
Calling pool registry with wagmi hooks and displaying pools from hook on Pool page.


src/constants/networks.ts
```diff
@@ -149,3 +165,135 @@ export const DEV_SUPPORTED_NETWORKS: SupportedNetworks = {
     blockExplorerUrls: ["https://explorer.evm-alpha.kava.io"],
   },
 }
+
+const evmosChain: Chain = {
```
ug02fast, line 169: Config for `wagmi` connection.

PR #1241, opened 2022-08-19
pool and token filtering on pool page

ug02fast:
image
filter on pool or token name


nkrulikovsky (approved):
LGTM!

src/pages/Pools.tsx
```diff
@@ -23,6 +24,8 @@ import { useSelector } from "react-redux"
 function Pools(): ReactElement | null {
   const { account, chainId } = useActiveWeb3React()
   const basicPools = useContext(BasicPoolsContext)
```
hammeiam, line 26: Would you mind getting rid of basicPools while you're adding expandedPools?

PR #1244, opened 2022-08-22
Fix hamburger menu

nkrulikovsky:
- Because of lazy loading of pages, the drawer doesn't close after opening new page. so moved `TopMenu` outside of `Suspense`.
- Also created `Pages` component and moved all page import to that component.

## Screenshot

https://user-images.githubusercontent.com/75422800/185997476-3e8c6392-be35-4af8-9066-e7a074feaaf7.mov

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -2,14 +2,7 @@ import "react-toastify/dist/ReactToastify.css"
 
 import { AppDispatch, AppState } from "../state"
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
-import React, {
-  ReactElement,
-  Suspense,
-  lazy,
-  useCallback,
-  useEffect,
-} from "react"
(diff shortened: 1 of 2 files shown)


hammeiam (approved):

ug02fast (commented):
can you call the file Routes instead? is it fine to kill Suspense?

src/pages/App.tsx
```diff
@@ -89,99 +71,47 @@ export default function App(): ReactElement {
   return (
     <QueryClientProvider client={queryClient}>
       <ReactQueryDevtools initialIsOpen />
-      <Suspense fallback={null}>
```
ug02fast, line 92: suspense is kil?

nkrulikovsky: I am using `Suspense` under the `TopMenu` to wrap all pages lazy imported.

ug02fast (approved):

PR #1254, opened 2022-09-07
add rewarder contract for arb frax usds metapool

ug02fast:
added condition to handle additional rewarder contracts for sperax


hammeiam (approved):
Screen Shot 2022-09-07 at 7 41 26 PM

PR #1256, opened 2022-09-08
Display minichef APRs

hammeiam:
- update `utils/minichef` to calculate the amount of lpToken staked in minichef for each pool
- update `usePoolData` to calculate minichef apr from amount staked, tvl, and sdl price
- update `PoolOverview` to display apr

### DIFF
--- a/src/hooks/usePoolData.ts
+++ b/src/hooks/usePoolData.ts
@@ -1,10 +1,15 @@
-import { A_PRECISION, PoolTypes } from "./../constants/index"
+import { A_PRECISION, BN_1E18, PoolTypes } from "./../constants/index"
 import { AprsContext, GaugeApr } from "./../providers/AprsProvider"
 import {
   Partners,
   getThirdPartyDataForPool,
 } from "../utils/thirdPartyIntegrations"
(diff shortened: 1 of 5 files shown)


ug02fast (approved):
Impl looks good. I didn't confirm math is correct.

src/hooks/usePoolData.ts
```diff
@@ -247,6 +254,20 @@ export default function usePoolData(name?: string): PoolDataHookReturnType {
           expandedPool.poolAddress
         ] || { oneDayVolume: null, apy: null, utilization: null }
 
+        let [amountStakedInMinichefUSD, minichefSDLApr] = [Zero, Zero]
```
nguyenaiden, line 257: Since each of these are only assigned once, I don't think we need to create a tuple. We can probably extract this logic into a helper function that takes in `(priceDataForPool, poolMinichefData, tokenPricesUSD.SDL)` and just return the `minichefSDLApr`. We can set `const minichefSDLApr = getMinichefSDLApr(priceDataForPool, poolMinichefData, tokenPricesUSD.SDL) || Zero` if we want a default value

nguyenaiden (approved):
1 nit, up to your discretion.

PR #1257, opened 2022-09-09
[WIP] Retrieve sidechain gauge data and render in piechart while on mainnet

nguyenaiden:
***Description***: As a part of the Cross-chain gauges feature, all gauge relative weights of all mainnet + sidechain gauges are to be displayed in the veSDL pie chart.
Note: Still WIP pending deployment of the remaining contracts + reliable testing strategy.
image

### DIFF
--- a/src/providers/SidechainGaugeProvider.tsx
+++ b/src/providers/SidechainGaugeProvider.tsx
@@ -1,0 +1,47 @@
+import React, { ReactElement, useEffect, useState } from "react"
+import {
+  SidechainGauges,
+  getSidechainGaugeData,
+} from "../utils/sidechainGauges"
+import {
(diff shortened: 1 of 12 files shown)


src/providers/SidechainGaugeProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+        chainId,
+        rootGaugeFactory,
+      )
+
+      setSidechainGauges(sidechainGauges)
+    }
+
+    void fetchSidechainGauges()
+  }, [chainId, library, account, rootGaugeFactory, gaugeController])
+
```
nkrulikovsky, line 41: I think using we can more easily  fetch `sidechainGauges` without useEffect using `react-query`.

PR #1260, opened 2022-09-13
Remove gaugeToPoolAddress call in Gauges state

nguyenaiden:
***Description***: In preparation for retrieving gauges while on side chain, we're aiming to reduce the reliance on `gaugeHelperContract` as it doesn't exist on side chain. We no longer need `gaugeHelperContract.gaugeToPoolAddress()` to match a gauge to a pool. We now use the `BasicPools` context and retrieve the information from there.  With the current state on mainnet, this will reduce 1 large multicall step.

### DIFF
--- a/src/utils/gauges.ts
+++ b/src/utils/gauges.ts
@@ -33,13 +33,13 @@ export type Gauge = {
   gaugeTotalSupply: BigNumber
   gaugeWeight: BigNumber
   lpTokenAddress: string
-  poolAddress: string | null
-  poolName: string | null
+  poolAddress: string | null | undefined
+  poolName: string | null | undefined
   gaugeRelativeWeight: BigNumber
   workingBalances: BigNumber
   workingSupply: BigNumber
   rewards: GaugeReward[]
-  gaugeName: string | null
+  gaugeName: string | null | undefined
   isKilled: boolean
 }
 
@@ -68,6 +68,11 @@ export type GaugeUserReward = {
 export type GaugeRewardUserData = Partial<{
   [gaugeAddress: string]: GaugeUserReward
 }>
+
+export type GaugePool = {
+  poolName: string
+  poolAddress: string
+}
 
 export const initialGaugesState: Gauges = {
   gaugeCount: 0,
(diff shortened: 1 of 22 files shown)


src/utils/gauges.ts
```diff
@@ -232,8 +238,9 @@ export async function getGaugeData(
           gaugeName: gaugeNames[index],
           lpTokenAddress,
           isKilled: gaugeKillStatuses[index] ?? false,
-          poolAddress: isValidPoolAddress && poolAddress ? poolAddress : null,
-          poolName: gaugePool?.poolName || null,
+          poolAddress:
+            isValidPoolAddress && pool?.poolAddress ? pool?.poolAddress : null,
```
ug02fast, line 242: does it make sense to early return here? you should also be able to do `poolAddress: isValidPoolAddress && pool?.poolAddress` to shorten logic

ug02fast (approved):
lgtm just one observation

src/utils/gauges.ts
```diff
@@ -198,19 +191,32 @@ export async function getGaugeData(
...
       (previousGaugeData, gaugeAddress, index) => {
         const lpTokenAddress = gaugeLpTokenAddresses[index]?.toLowerCase()
-        const poolAddress = gaugePoolAddresses[index]
+        const pool = lpTokenToPool[lpTokenAddress]
         const isValidPoolAddress = Boolean(
-          poolAddress && !isAddressZero(poolAddress),
-        )
-        const gaugePool = Object.values(basicPools || {}).find(
-          (pool) => pool.poolAddress === poolAddress,
+          pool && pool.poolAddress && !isAddressZero(pool.poolAddress),
```
ug02fast, line 219: you should be able to shorten this to `pool?.poolAddress && ...`

hammeiam: agreed, I don't think poolAddress would ever be 0x0 here

hammeiam: Furthermore since you just created that map and you know that each key has a valid value, you can just say `const pool = lpTokenToPool[lpTokenAddress] as MyPoolType` so you don't have to do the null checks later on

src/utils/gauges.ts
hammeiam: agreed

src/utils/gauges.ts
nguyenaiden: @ug02fast, I'm running into the issue of having to check for both `undefined` as well as `null` types. This is mainly coming from our need to use `?` optional chaining, but the `.tryAll` method is returning `Promise<null>` if the call fails.

What do you think about checking for both, or we pick 1 type of falsy value and go with that?

ug02fast (approved):

PR #1261, opened 2022-09-13
fix broken saddle logo in MetaMask

ug02fast:
This should be a decent solution until we support adding other tokens to MM.

### DIFF
--- a/src/hooks/useAddTokenToMetamask.ts
+++ b/src/hooks/useAddTokenToMetamask.ts
@@ -1,7 +1,7 @@
 import { SUPPORTED_WALLETS, Token } from "../constants"
 import { useCallback, useState } from "react"
-import { find } from "lodash"
 
+import { find } from "lodash"
 import { getTokenIconPath } from "../utils"
 import { useActiveWeb3React } from "./index"
 
@@ -30,7 +30,10 @@ export default function useAddTokenToMetamask(token: Token | undefined): {
               address: token.addresses[chainId],
               symbol: token.symbol,
               decimals: token.decimals,
-              image: getTokenIconPath(token.symbol),
+              image:
+                token.symbol === "SDL"
+                  ? "https://assets.coingecko.com/coins/images/20476/small/edit_saddle.png?1660734429"
+                  : getTokenIconPath(token.symbol),
             },
           },
         })


src/hooks/useAddTokenToMetamask.ts
```diff
@@ -30,7 +30,10 @@ export default function useAddTokenToMetamask(token: Token | undefined): {
               address: token.addresses[chainId],
               symbol: token.symbol,
               decimals: token.decimals,
-              image: getTokenIconPath(token.symbol),
+              image:
+                token.symbol === "SDL"
+                  ? "https://assets.coingecko.com/coins/images/20476/small/edit_saddle.png?1660734429"
+                  : getTokenIconPath(token.symbol),
```
nkrulikovsky, line 36: LGTM!

This logo is better than the logo of saddle token on sushiswap.
For some reason sushi swap saddle logo is having dark border.

nkrulikovsky (approved):

src/hooks/useAddTokenToMetamask.ts
```diff
@@ -30,7 +30,10 @@ export default function useAddTokenToMetamask(token: Token | undefined): {
               address: token.addresses[chainId],
               symbol: token.symbol,
               decimals: token.decimals,
-              image: getTokenIconPath(token.symbol),
+              image:
+                token.symbol === "SDL"
+                  ? "https://assets.coingecko.com/coins/images/20476/small/edit_saddle.png?1660734429"
```
nguyenaiden, line 35: Can we save this as a static variable up at the top somewhere?
const SDL_TOKEN_ICON_LINK = "https://assets.coingecko.com/coins/images/20476/small/edit_saddle.png?1660734429"

nguyenaiden (approved):
+1

src/hooks/useAddTokenToMetamask.ts
hammeiam, line 38: I think it would be better if we just fix this at the source by making that helper function return a url rather than relative path, eg:

```typescript
export function getTokenIconPath(tokenSymbol: string): string {
  const iconName = tokenSymbol.toLowerCase().includes("saddle")
    ? "saddle_lp_token"
    : tokenSymbol.toLowerCase()
  const x = new Image()
  x.src = `/static/icons/svg/${iconName}.svg`

  return x.src
}
```

Alternatively, I think just doing `${window.location.host}/static/icons/svg/${iconName}.svg` should work as well

nguyenaiden: Actually, gonna +1 on this comment.  This is a better solution.

ug02fast: 🐐

PR #1267, opened 2022-09-22
Address Registry Provider

nguyenaiden:
***Description***: The Address Registry Provider is responsible for calling the MasterRegistry and resolve addresses that are used often. This is also the start of moving away from accessing contracts as hooks to instantiating `Contract` objects whenever they're needed.

The first contract address is `ChildGaugeFactory`. Only available on non-mainnet/hardhat chainIds.

### DIFF
--- a/src/providers/RegistryAddressProvider.tsx
+++ b/src/providers/RegistryAddressProvider.tsx
@@ -1,0 +1,69 @@
+import {
+  CHILD_GAUGE_FACTORY_NAME,
+  useMasterRegistry,
+} from "../hooks/useContract"
+import {
+  EMPTY_LOADABLE,
+  LoadableType,
+  useLoadingState,
+} from "../utils/loadable"
+import React, { ReactElement, useEffect } from "react"
+import { ChainId } from "../constants"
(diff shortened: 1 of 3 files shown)


src/providers/RegistryAddressProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+            )
+          addresses[parseBytes32String(CHILD_GAUGE_FACTORY_NAME)] =
+            childGaugeFactoryAddress
+          onSuccess(addresses)
+        }
+      } catch (error) {
+        const errorMessage =
+          "Unable to retrieve and set addresses from MasterRegistry"
+        console.error(errorMessage)
+        onFailure(error as Error)
```
lionpunk, line 59: How can use this error to show on toast or on some other else?

nkrulikovsky, line 59: How can we show this error message on toast if we want to display for users?

nguyenaiden:
Closing per https://github.com/saddle-finance/saddle-frontend/pull/1268/files

PR #1268, opened 2022-09-22
Address Registry React Query Hook

nguyenaiden:
***Description***: Resolve contract addresses with the MasterRegistry via a hook using ReactQuery
Related PR: https://github.com/saddle-finance/saddle-frontend/pull/1267


src/providers/useRegistryAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type RegistryAddress = Partial<Record<string, string>>
+
+export const useRegistryAddress = (): UseQueryResult<RegistryAddress> => {
+  const { chainId } = useActiveWeb3React()
+  const masterRegistry = useMasterRegistry()
+
+  return useQuery(["registryAddress"], async () => {
+    if (!masterRegistry) {
+      console.error(
```
ug02fast, line 18: You can throw error here and rq will handle it correctly.

ug02fast (commented):
Lgtm just one small thing

src/providers/useRegistryAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+        "Master Registry not available. Unable to retrieve contract addresses",
+      )
+      return {}
```
ug02fast, line 21:
```suggestion
      threw new Error("Master Registry not available. Unable to retrieve contract addresses")
```

ug02fast (commented):
suggestion

ug02fast (approved):

src/providers/useRegistryAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (!masterRegistry) {
+      throw new Error("Master Registry not available. Unable to retrieve contract addresses")
+    }
+
+    if (chainId !== ChainId.MAINNET && chainId !== ChainId.HARDHAT) {
+      const childGaugeFactoryAddress =
+        await masterRegistry.resolveNameToLatestAddress(
+          CHILD_GAUGE_FACTORY_NAME,
+        )
+      const addresses: Partial<Record<string, string>> = {}
```
hammeiam, line 26: We should hoist this outside of the if block, then return it eg

```
const addr = {}
if (){
 // maybe add stuff to addr
}
return addr
```

src/providers/useRegistryAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+      addresses[parseBytes32String(CHILD_GAUGE_FACTORY_NAME)] =
```
hammeiam, line 27: thoughts on changing `CHILD_GAUGE_FACTORY_NAME` to `CHILD_GAUGE_FACTORY_NAME_BYTES32` or storing it as the parsed name and only formatting it for the contract call?

src/providers/useRegistryAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+import { parseBytes32String } from "ethers/lib/utils"
+import { useActiveWeb3React } from "../hooks"
+
+export type RegistryAddress = Partial<Record<string, string>>
+
+export const useRegistryAddress = (): UseQueryResult<RegistryAddress> => {
+  const { chainId } = useActiveWeb3React()
+  const masterRegistry = useMasterRegistry()
+
+  return useQuery(["registryAddress"], async () => {
```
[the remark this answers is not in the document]
nguyenaiden: It can, but we can use multiCall whenever we have more addresses.

src/providers/useRegistryAddress.ts
nguyenaiden: +1 on this.

src/providers/useRegistryAddress.ts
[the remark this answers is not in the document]
hammeiam: And if so, would `react-query` prefer to manage each of those requests independently? I'm not sure.

hammeiam, line 16: Also reminder to add with a map or enum of react query keys to avoid misspellings

hammeiam (approved):
lgtm!

PR #1287, opened 2022-10-12
Cross chain Gauges

nguyenaiden:
***Description***: In preparation for crosschain gauges, the `Gauges` component along with `GaugeProvider` have been updated to be able to fully build the `Gauge` object regardless of which chain the user is on. 

Features added:
- Staking/Unstaking/Claiming on sidechain gauges
- gauge vote piechart now includes sidechain gauges as well

*Mainnet vs Sidechain*
```
LiquidityGaugeV5 vs. ChildGauge (gauge specific calls)
GaugeController vs ChildGaugeFactory (to retrieve gauge counts)
GaugeMinterContract vs ChildGaugeFactory (rewards minting)
```

Gated behind feature flag.
Feature pic with dummy gauge data on Arbitrum and Optimism.

Screen Shot 2022-11-11 at 12 04 39 PM

### DIFF
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -1,6 +1,8 @@
 import {
   BRIDGE_CONTRACT_ADDRESSES,
   BTC_POOL_NAME,
+  CHILD_ORACLE_CONTRACT_ADDRESSES,
+  ChainId,
   FEE_DISTRIBUTOR_ADDRESSES,
   GAUGE_CONTROLLER_ADDRESSES,
   GAUGE_MINTER_ADDRESSES,
@@ -23,6 +25,12 @@ import { AddressZero } from "@ethersproject/constants"
 import BRIDGE_CONTRACT_ABI from "../constants/abis/bridge.json"
 import { BasicPoolsContext } from "../providers/BasicPoolsProvider"
 import { Bridge } from "../../types/ethers-contracts/Bridge"
+import CHILD_GAUGE_ABI from "../constants/abis/childGauge.json"
+import CHILD_GAUGE_FACTORY_ABI from "../constants/abis/childGaugeFactory.json"
+import CHILD_ORACLE_ABI from "../constants/abis/childOracle.json"
+import { ChildGauge } from "../../types/ethers-contracts/ChildGauge"
+import { ChildGaugeFactory } from "../../types/ethers-contracts/ChildGaugeFactory"
+import { ChildOracle } from "../../types/ethers-contracts/ChildOracle"
 import ERC20_ABI from "../constants/abis/erc20.json"
 import FEE_DISTRIBUTOR_ABI from "../constants/abis/feeDistributor.json"
 import { FeeDistributor } from "../../types/ethers-contracts/FeeDistributor"
@@ -64,6 +72,7 @@ import { SynthetixExchangeRate } from "../../types/ethers-contracts/SynthetixExc
 import { SynthetixNetworkToken } from "../../types/ethers-contracts/SynthetixNetworkToken"
 import VOTING_ESCROW_CONTRACT_ABI from "../constants/abis/votingEscrow.json"
 import { VotingEscrow } from "../../types/ethers-contracts/VotingEscrow"
+import { Web3Provider } from "@ethersproject/providers"
 import { formatBytes32String } from "@ethersproject/strings"
 import { useActiveWeb3React } from "./index"
 
@@ -390,4 +399,71 @@ export function useLiquidityGaugeContract(
   gaugeAddress?: string,
 ): LiquidityGaugeV5 | null {
   return useContract(gaugeAddress, LIQUIDITY_V5_GAUGE_ABI) as LiquidityGaugeV5
+}
+
+// This section instantiate new contract without Hooks
+export function getGaugeContract(
+  library: Web3Provider,
+  chainId: ChainId,
+  address: string,
+  account: string,
+): LiquidityGaugeV5 | ChildGauge {
+  if (chainId === ChainId.MAINNET) {
+    return getContract(
+      address,
+      LIQUIDITY_V5_GAUGE_ABI,
+      library,
+      account,
+    ) as LiquidityGaugeV5
+  }
+
+  return getContract(address, CHILD_GAUGE_ABI, library, account) as ChildGauge
+}
+
+export function getGaugeMinterContract(
+  chainId: ChainId,
+  account: string,
+  library: Web3Provider,
+) {
+  return getContract(
+    GAUGE_MINTER_ADDRESSES[chainId],
+    GAUGE_MINTER_ABI,
+    library,
+    account,
+  ) as Minter
+}
+
+export function getChildGaugeFactory(library: Web3Provider, address: string) {
+  if (!address) return null
+  return getContract(
+    address,
+    CHILD_GAUGE_FACTORY_ABI,
+    library,
+  ) as ChildGaugeFactory
+}
+
+export const getVotingEscrowContract = (
+  chainId: ChainId,
+  account: string,
+  library: Web3Provider,
+): VotingEscrow => {
+  return getContract(
+    VOTING_ESCROW_CONTRACT_ADDRESS[chainId],
+    VOTING_ESCROW_CONTRACT_ABI,
+    library,
+    account,
+  ) as VotingEscrow
+}
+
+export const getChildOracle = (
+  chainId: ChainId,
+  account: string,
+  library: Web3Provider,
+): ChildOracle => {
+  return getContract(
+    CHILD_ORACLE_CONTRACT_ADDRESSES[chainId],
+    CHILD_ORACLE_ABI,
+    library,
+    account,
+  ) as ChildOracle
 }
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -421,8 +421,8 @@ export function getGaugeContract(
 }
 
 export function getGaugeControllerContract(
-  chainId: ChainId,
   library: Web3Provider,
+  chainId: ChainId,
   account?: string,
 ) {
   return getContract(
@@ -434,8 +434,8 @@ export function getGaugeControllerContract(
 }
 
 export function getGaugeMinterContract(
-  chainId: ChainId,
   library: Web3Provider,
+  chainId: ChainId,
   account?: string,
 ) {
   return getContract(
@@ -446,18 +446,24 @@ export function getGaugeMinterContract(
   ) as Minter
 }
 
-export function getChildGaugeFactory(library: Web3Provider, address: string) {
+export function getChildGaugeFactory(
+  library: Web3Provider,
+  chainId: ChainId,
+  address: string,
+  account?: string,
+) {
   return getContract(
     address,
     CHILD_GAUGE_FACTORY_ABI,
     library,
+    account ?? undefined,
   ) as ChildGaugeFactory
 }
 
 export const getVotingEscrowContract = (
-  chainId: ChainId,
-  account: string,
   library: Web3Provider,
+  chainId: ChainId,
+  account?: string,
 ): VotingEscrow => {
   return getContract(
     VOTING_ESCROW_CONTRACT_ADDRESS[chainId],
@@ -468,9 +474,9 @@ export const getVotingEscrowContract = (
 }
 
 export const getChildOracle = (
-  chainId: ChainId,
-  account: string,
   library: Web3Provider,
+  chainId: ChainId,
+  account?: string,
 ): ChildOracle => {
   return getContract(
     CHILD_ORACLE_CONTRACT_ADDRESSES[chainId],
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -449,11 +449,11 @@ export const getGaugeMinterContract = (
 export const getChildGaugeFactory = (
   library: Web3Provider,
   chainId: ChainId,
-  address: string,
+  address?: string,
   account?: string,
 ) => {
   return getContract(
-    address,
+    address ?? "",
     CHILD_GAUGE_FACTORY_ABI,
     library,
     account ?? undefined,
@@ -486,6 +486,8 @@ export const getChildOracle = (
   ) as ChildOracle
 }
 
-export const isMainnet = (chainId: ChainId) => {
-  return chainId === ChainId.MAINNET || chainId === ChainId.HARDHAT
+export const isMainnet = (chainId?: ChainId): boolean => {
+  return (
+    !!chainId && (chainId === ChainId.MAINNET || chainId === ChainId.HARDHAT)
+  )
 }
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -402,12 +402,12 @@ export function useLiquidityGaugeContract(
 }
 
 // This section instantiate new contract without Hooks
-export function getGaugeContract(
+export const getGaugeContract = (
   library: Web3Provider,
   chainId: ChainId,
   address: string,
   account: string,
-): LiquidityGaugeV5 | ChildGauge {
+): LiquidityGaugeV5 | ChildGauge => {
   if (chainId === ChainId.MAINNET) {
     return getContract(
       address,
@@ -420,11 +420,11 @@ export function getGaugeContract(
   return getContract(address, CHILD_GAUGE_ABI, library, account) as ChildGauge
 }
 
-export function getGaugeControllerContract(
+export const getGaugeControllerContract = (
   library: Web3Provider,
   chainId: ChainId,
   account?: string,
-) {
+) => {
   return getContract(
     GAUGE_CONTROLLER_ADDRESSES[chainId],
     GAUGE_CONTROLLER_ABI,
@@ -433,11 +433,11 @@ export function getGaugeControllerContract(
   ) as GaugeController
 }
 
-export function getGaugeMinterContract(
+export const getGaugeMinterContract = (
   library: Web3Provider,
   chainId: ChainId,
   account?: string,
-) {
+) => {
   return getContract(
     GAUGE_MINTER_ADDRESSES[chainId],
     GAUGE_MINTER_ABI,
@@ -446,12 +446,12 @@ export function getGaugeMinterContract(
   ) as Minter
 }
 
-export function getChildGaugeFactory(
+export const getChildGaugeFactory = (
   library: Web3Provider,
   chainId: ChainId,
   address: string,
   account?: string,
-) {
+) => {
   return getContract(
     address,
     CHILD_GAUGE_FACTORY_ABI,
@@ -484,4 +484,8 @@ export const getChildOracle = (
     library,
     account,
   ) as ChildOracle
+}
+
+export const isMainnet = (chainId: ChainId) => {
+  return chainId === ChainId.MAINNET || chainId === ChainId.HARDHAT
 }
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -10,6 +10,7 @@ import {
   MASTER_REGISTRY_CONTRACT_ADDRESSES,
   MINICHEF_CONTRACT_ADDRESSES,
   RETROACTIVE_VESTING_CONTRACT_ADDRESSES,
+  ROOT_GAUGE_FACTORY_CONTRACT_ADDRESSES,
   SDL_TOKEN_ADDRESSES,
   SDL_WETH_SUSHI_LP_CONTRACT_ADDRESSES,
   SYNTHETIX_CONTRACT_ADDRESSES,
@@ -58,7 +59,9 @@ import POOL_REGISTRY_ABI from "../constants/abis/poolRegistry.json"
 import { PermissionlessDeployer } from "../../types/ethers-contracts/PermissionlessDeployer"
 import { PoolRegistry } from "../../types/ethers-contracts/PoolRegistry"
 import RETROACTIVE_VESTING_CONTRACT_ABI from "../constants/abis/retroactiveVesting.json"
+import ROOT_GAUGE_FACTORY_ABI from "../constants/abis/rootGaugeFactory.json"
 import { RetroactiveVesting } from "../../types/ethers-contracts/RetroactiveVesting"
+import { RootGaugeFactory } from "../../types/ethers-contracts/RootGaugeFactory"
 import SDL_TOKEN_ABI from "../constants/abis/sdl.json"
 import SUSHI_POOL_ABI from "../constants/abis/sushiPool.json"
 import SYNTHETIX_EXCHANGE_RATE_CONTRACT_ABI from "../constants/abis/synthetixExchangeRate.json"
@@ -484,6 +487,19 @@ export const getChildOracle = (
     library,
     account,
   ) as ChildOracle
+}
+
+export const getRootGaugeFactory = (
+  library: Web3Provider,
+  chainId: ChainId,
+  account?: string,
+): RootGaugeFactory => {
+  return getContract(
+    ROOT_GAUGE_FACTORY_CONTRACT_ADDRESSES[chainId],
+    ROOT_GAUGE_FACTORY_ABI,
+    library,
+    account,
+  ) as RootGaugeFactory
 }
 
 export const isMainnet = (chainId?: ChainId): boolean => {
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -420,10 +420,23 @@ export function getGaugeContract(
   return getContract(address, CHILD_GAUGE_ABI, library, account) as ChildGauge
 }
 
+export function getGaugeControllerContract(
+  chainId: ChainId,
+  library: Web3Provider,
+  account?: string,
+) {
+  return getContract(
+    GAUGE_CONTROLLER_ADDRESSES[chainId],
+    GAUGE_CONTROLLER_ABI,
+    library,
+    account,
+  ) as GaugeController
+}
+
 export function getGaugeMinterContract(
   chainId: ChainId,
-  account: string,
   library: Web3Provider,
+  account?: string,
 ) {
   return getContract(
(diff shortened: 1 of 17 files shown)


src/utils/gauges.ts
```diff
@@ -15,10 +24,12 @@ import {
 import { BasicPools } from "./../providers/BasicPoolsProvider"
 import { BasicToken } from "./../providers/TokensProvider"
 import { BigNumber } from "@ethersproject/bignumber"
+import CHILD_GAUGE_ABI from "../constants/abis/childGauge.json"
```
nkrulikovsky, line 27: Could we move `CHILD_GAUGE_ABI` to `constants` folder because it is using on `useContract` as well?

nkrulikovsky (commented):
If we need to use `CHILD_GAUGE_ABI` on several files, could we move it to constants folder?

src/utils/gauges.ts
```diff
@@ -82,62 +98,88 @@ export const initialGaugesState: Gauges = {
...
-      ),
-    )
+    const gaugeWeightsPromise: Promise<BigNumber[]> =
+      chainId === ChainId.MAINNET
+        ? ethCallProvider.all(
+            gaugeAddresses.map((gaugeAddress) =>
+              gaugeControllerMultiCall.get_gauge_weight(gaugeAddress),
+            ),
+          )
+        : Promise.resolve(Array<BigNumber>(gaugeAddresses.length).fill(Zero))
```
penandlim, line 171: ✅

src/utils/gauges.ts
```diff
@@ -82,62 +98,88 @@ export const initialGaugesState: Gauges = {
...
+    const gaugeRelativeWeightsPromise: Promise<BigNumber[]> =
+      chainId === ChainId.MAINNET
+        ? ethCallProvider.all(
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+            gaugeAddresses.map((gaugeAddress) =>
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call
+              gaugeControllerMultiCall.gauge_relative_weight(gaugeAddress),
+            ),
+          )
+        : Promise.resolve(Array<BigNumber>(gaugeAddresses.length).fill(Zero))
```
penandlim, line 182: Looks good for mainnet calls.
Just wondering whats the implication of being on sidechain and this promise returning an array with 0s?

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  childGaugeFactory: ChildGaugeFactory | null,
+  gaugeController: GaugeController | null,
+): Promise<BigNumber> {
+  if (childGaugeFactory) {
+    return await childGaugeFactory.get_gauge_count()
+  }
+
+  if (gaugeController) {
+    return await gaugeController.n_gauges()
+  }
```
penandlim, line 444: Heads up: on a side chain `childGaugeFactory.get_gauge_count()` will return the number of ChildGauges existing on the side chain. Including permissionlessly deployed ChildGauges. (any user can trigger a deployment of a childgauge via `ChildGaugeFactory` but the corresponding root gauge will not be included in gauge controller until its been added by our multisig)

But on mainnet, `gaugeController.n_gauges()` returns number of RootGauges + LiquidityGaugeV5s that are approved by Saddle DAO.

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+      createMultiCallContract<LiquidityGaugeV5>(
+        address,
+        LIQUIDITY_GAUGE_V5_ABI,
+      ),
+    )
+  }
+
+  return gaugeAddresses.map((address) =>
+    createMultiCallContract<ChildGauge>(address, CHILD_GAUGE_ABI),
+  )
```
penandlim, line 481: ✅

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  gaugeRelativeWeights: BigNumber[],
+  gaugeMulticallContracts: MulticallContract<ChildGauge>[],
+  ethCallProvider: MulticallProvider,
+) {
+  if (gaugeMinterContract) {
+    const minterSDLRate = await gaugeMinterContract.rate()
+    return gaugeRelativeWeights.map((weight) =>
+      minterSDLRate.mul(weight).div(BN_1E18),
+    )
+  }
```
penandlim, line 495: ✅

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+
+  const currentTimeStamp = Date.now() / 1000
+  return await ethCallProvider.tryAll(
+    gaugeMulticallContracts.map((contract) =>
+      contract.inflation_rate(Math.floor(currentTimeStamp / 604800)),
+    ),
+  )
```
penandlim, line 502: ✅, 

Fetches current epoch's inflation rate of each child gauge.
If the chain is sped up or out of sync, this could return 0 or an old value.

In production setting, we expect this to happen for a brief moment at the beginning of every new epoch. After an epoch passes, an interaction needs to happen with this contract in order to advance the epoch counter. Therefore couple users may see 0 until someone interacts with it (checkpoint, claim SDL, claim exteranl rewards, deposit, unstake, etc).

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  gaugeMulticallContracts: (
+    | MulticallContract<ChildGauge>
+    | MulticallContract<LiquidityGaugeV5>
+  )[],
+  gaugeRewardCounts: BigNumber[],
+  ethCallProvider: MulticallProvider,
+) {
+  return Promise.all(
+    gaugeRewardCounts.map((count, index) =>
+      ethCallProvider.all(
```
penandlim, line 528: Should this be tryAll?

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  gaugeMulticallContracts:
+    | MulticallContract<LiquidityGaugeV5>[]
+    | MulticallContract<ChildGauge>[],
+  ethCallProvider: MulticallProvider,
+) {
+  return (
+    await ethCallProvider.tryAll(
+      gaugeMulticallContracts.map((contract) => contract.reward_count()),
+    )
+  ).map((count) => count || Zero)
```
penandlim, line 515: ✅

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+    | MulticallContract<LiquidityGaugeV5>
+  )[],
+  gaugeRewardCounts: BigNumber[],
+  ethCallProvider: MulticallProvider,
+) {
+  return Promise.all(
+    gaugeRewardCounts.map((count, index) =>
+      ethCallProvider.all(
+        enumerate(count.toNumber(), 0).map((num) =>
+          gaugeMulticallContracts[index].reward_tokens(num),
```
penandlim, line 530: ✅

penandlim: Heads up. Since RootGauge lacks most of these functions relating to rewards, these calls will fail.

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+              [arg0: string, overrides?: CallOverrides | undefined],
+              [string, string, BigNumber, BigNumber, BigNumber, BigNumber] & {
+                token: string
+                distributor: string
+                period_finish: BigNumber
+                rate: BigNumber
+                last_update: BigNumber
+                integral: BigNumber
+              }
+            >,
```
penandlim, line 563: `ChildGauge.reward_data(address)` will return a different struct

```
struct Reward:
    distributor: address
    period_finish: uint256
    rate: uint256
    last_update: uint256
    integral: uint256
```

penandlim: orr maybe typescript is smart enough to fill in 0 for the missing key?

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+            gaugeMulticallContracts[index].claimable_reward(
+              account,
+              gaugeRewardsTokens[index][num],
+            ) as unknown as MulticallCall<
+              [
+                _user: string,
+                _reward_token: string,
+                overrides?: CallOverrides | undefined,
+              ],
+              BigNumber
```
penandlim, line 594: ✅

src/utils/gauges.ts
nguyenaiden: @penandlim  Since we're not allowing user to view the veSDL page while on side chains, we don't need to populate the weights. Will go through it again to see if we need the weights and relative weights for anything else.

src/utils/gauges.ts
nguyenaiden: @penandlim if that's the case, is there a flag or some call we can make to ensure that we only display gauges that have been added by our multisig?

src/utils/gauges.ts
nguyenaiden: Good catch. Will update.

src/utils/gauges.ts
nguyenaiden: Hmm, gotta go back to the drawing board to see how I could handle `RootGauges`  and `LiquidityV5Gauges`.  we would only want to calculate rewards for mainnet gauges since we don't allow users to claim sidechain gauges (represented by rootGauges) while on mainnet.

Looking for a way to discern between the 2 types of gauges.

src/utils/gauges.ts
nguyenaiden: iirc, the value comes back as either 0 or null. Either way, it hasn't caused any compatibility issues.

src/utils/gauges.ts
nguyenaiden: Hmm I don't want to move abis to a different place. Was that what you meant? Unless I'm misinterpreting your message?

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+}
+
+function retrieveGaugeAddresses(
+  gaugeControllerMultiCall: MulticallContract<GaugeController>,
+  childGaugeFactoryMultiCall: MulticallContract<ChildGaugeFactory>,
+): (
+  value: number,
+  index: number,
+  array: number[],
+) => MulticallCall<
```
hammeiam, line 456: Protip:
`ReturnType<MulticallContract<GaugeController>["gauges"]>`

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  }
+
+  if (gaugeController) {
+    return await gaugeController.n_gauges()
+  }
+
+  return Promise.resolve(Zero)
+}
+
+function retrieveGaugeAddresses(
```
hammeiam, line 449: seeing overloading comment above

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  ethCallProvider: MulticallProvider,
+) {
+  if (gaugeMinterContract) {
+    const minterSDLRate = await gaugeMinterContract.rate()
+    return gaugeRelativeWeights.map((weight) =>
+      minterSDLRate.mul(weight).div(BN_1E18),
+    )
+  }
+
+  const currentTimeStamp = Date.now() / 1000
```
hammeiam, line 497: May as well do the `Math.floor` here instead of inline

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+  return await ethCallProvider.tryAll(
+    gaugeMulticallContracts.map((contract) =>
+      contract.inflation_rate(Math.floor(currentTimeStamp / 604800)),
```
hammeiam, line 500: Comment or const for what `604800` is

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+      ),
+    )
+  }
+
+  return gaugeAddresses.map((address) =>
+    createMultiCallContract<ChildGauge>(address, CHILD_GAUGE_ABI),
+  )
+}
+
+async function getSDLRates(
```
hammeiam, line 484: seeing overloading comment above

src/utils/gauges.ts
hammeiam: At some point we should start using `chunkedTryAll` everywhere to avoid the case where we are trying to make more many calls in one operation than multicall can handle. See `useBasicTokens` for example

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+async function getGaugeRewardsFromTokens(
+  gaugeRewardCounts: BigNumber[],
+  gaugeMulticallContracts: (
+    | MulticallContract<ChildGauge>
+    | MulticallContract<LiquidityGaugeV5>
+  )[],
+  gaugeRewardsTokens: string[][],
+  ethCallProvider: MulticallProvider,
+) {
+  return Promise.all(
```
hammeiam, line 546: Kinda cool seeing this `all(map(all(map`

src/utils/gauges.ts
```diff
@@ -372,3 +430,171 @@ export function areGaugesActive(chainId?: ChainId): boolean {
...
+    gaugeRewardCounts.map((count, index) =>
+      ethCallProvider.all(
+        enumerate(count.toNumber(), 0).map(
+          (num) =>
+            gaugeMulticallContracts[index].reward_data(
+              gaugeRewardsTokens[index][num],
+            ) as unknown as MulticallCall<
```
hammeiam, line 553: Doesn't TS infer this type?

src/utils/gauges.ts
hammeiam: My man's trying to make csv work when parquet might be a cleaner choice. `=` vs `||`

src/utils/gauges.ts
penandlim: Unfortunately not on side chains :/

The official list of gauge addresses that are eligible to be voted by veSDL holders is maintained in GaugeController contract on eth mainnet.

So would have to rely on an offchain or cross chain solution.

Off chain API would be the best imo.

src/utils/gauges.ts
penandlim: https://github.com/saddle-finance/saddle-contract/blob/923e8a398bcdf5c1144e214f883f5f27b9f525cd/contracts/xchainGauges/implementations/RootGauge.vy#L47-L49

One solution is to use `tryAll()` and attempt to read one of these values. 

Only root gauges will return valid values while liquidity v5 will be reverted (returns null on multicall)

src/utils/gauges.ts
penandlim: Safe to ignore this as ChildGaugeFactory has unique LPToken <> Gauge mapping

Can be overriden by the multisig in case malicious user.

Although we cannot check whether this child gauge is added to the gauge controller yet, we can safely assume it is if it's SDL emission is >0. 

When SDL emission is 0, it would mean no one voted for the corresponding root gauge or that the root gauge is not added to the gauge controller yet.

src/hooks/useContract.ts
```diff
@@ -391,3 +400,70 @@ export function useLiquidityGaugeContract(
...
+  return getContract(
+    GAUGE_MINTER_ADDRESSES[chainId],
+    GAUGE_MINTER_ABI,
+    library,
+    account,
+  ) as Minter
+}
+
+export function getChildGaugeFactory(library: Web3Provider, address: string) {
+  if (!address) return null
```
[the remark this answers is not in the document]
nguyenaiden: Ack. I'll add this to all of them.

src/hooks/useContract.ts
```diff
@@ -391,3 +400,70 @@ export function useLiquidityGaugeContract(
...
+    VOTING_ESCROW_CONTRACT_ADDRESS[chainId],
+    VOTING_ESCROW_CONTRACT_ABI,
+    library,
+    account,
+  ) as VotingEscrow
+}
+
+export const getChildOracle = (
+  chainId: ChainId,
+  account: string,
```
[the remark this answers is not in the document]
nguyenaiden: @penandlim This is what I was referring to. I agree with David that we wouldn't need to pass in `account` here as optional. But for consistency's sake, we should pass in `account` if it's available everywhere. Thoughts?

src/hooks/useContract.ts
```diff
@@ -391,3 +400,70 @@ export function useLiquidityGaugeContract(
...
+  library: Web3Provider,
+): ChildOracle => {
+  return getContract(
```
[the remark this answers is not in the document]
nguyenaiden: I'll keep this in mind as I go through the refactor. It might not even be necessary to have this check because if we wrap everything in a try/catch, we can have the `GaugeProvider` return null for gauges if any errors come up. 

I'm leaning towards letting `getContract` throw an error, and have the providers handle what and when to return null.

src/utils/gauges.ts
```diff
@@ -82,62 +98,88 @@ export const initialGaugesState: Gauges = {
 export async function getGaugeData(
   library: Web3Provider,
   chainId: ChainId,
-  gaugeController: GaugeController,
   basicPools: BasicPools,
-  gaugeMinterContract: Minter,
+  gaugeController: GaugeController | null,
+  gaugeMinterContract: Minter | null,
+  registryAddresses: Partial<Record<string, string>>,
```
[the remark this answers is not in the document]
nguyenaiden: Ack. will add this in the next commit

src/hooks/useUserGauge.ts
```diff
@@ -95,11 +111,35 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
...
+        if (registryAddresses["ChildGaugeFactory"]) {
+          const childGaugeFactory = getChildGaugeFactory(
+            library,
+            chainId,
+            registryAddresses["ChildGaugeFactory"],
+            account,
+          )
+          promises.push(childGaugeFactory.mint(gaugeAddress))
+        } else {
+          console.error(
```
hammeiam, line 137: We should display this err to user

src/hooks/useUserGauge.ts
```diff
@@ -111,3 +151,28 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
...
+  setVeSdlBalance: (value: SetStateAction<BigNumber>) => void,
+  setTotalVeSdl: (value: SetStateAction<BigNumber>) => void,
+): Promise<void> {
+  if (chainId === ChainId.MAINNET || chainId === ChainId.HARDHAT) {
+    const votingEscrowContract = getVotingEscrowContract(
+      library,
+      chainId,
+      account,
+    )
+    setVeSdlBalance(await votingEscrowContract["balanceOf(address)"](account))
```
hammeiam, line 168: can parallelize these awaits

src/hooks/useUserGauge.ts
```diff
@@ -111,3 +151,28 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
...
+    const supply = await votingEscrowContract["totalSupply()"]()
+    setTotalVeSdl(supply)
+  } else {
+    const childOracle = getChildOracle(library, chainId, account)
+    const veSDLBalance = await childOracle.balanceOf(account)
```
hammeiam, line 173: ditto

src/pages/Farm/ClaimRewardsDlg.tsx
```diff
@@ -32,8 +32,8 @@ export default function ClaimRewardsDlg({
   gaugeAddress,
   displayName,
 }: Props): JSX.Element {
-  const { chainId } = useActiveWeb3React()
-  const userGauge = useUserGauge(gaugeAddress)
+  const { account, library, chainId } = useActiveWeb3React()
+  const userGauge = useUserGauge(account, chainId, library, gaugeAddress)
```
hammeiam, line 36: prefer to access context values via hooks instead of via props

src/providers/UserStateProvider.tsx
```diff
@@ -68,8 +68,8 @@ export default function UserStateProvider({
       )
       const gaugeRewardsPromise = gauges
         ? getGaugeRewardsUserData(
-            library,
             chainId,
+            library,
```
hammeiam, line 72: wai

src/utils/gauges.ts
```diff
@@ -27,20 +36,25 @@ import { Web3Provider } from "@ethersproject/providers"
 import { Zero } from "@ethersproject/constants"
 import { isAddressZero } from "."
 
-export type Gauge = {
+export type Gauge = BaseGauge & GaugeWeight
+
+export type GaugeWeight = {
```
hammeiam, line 41: add a comment about the diff btw mainnet and sidechain gauges

src/utils/gauges.ts
```diff
@@ -82,252 +105,263 @@ export const initialGaugesState: Gauges = {
...
+/* ------- Start of helper functions ------- */
+// Mainnet specific
+async function buildGaugeData(
+  library: Web3Provider,
+  chainId: ChainId,
+  basicPools: BasicPools,
+  registryAddresses: Partial<Record<string, string>>,
+  account?: string,
+) {
+  if (!registryAddresses || !areGaugesActive(chainId)) return initialGaugesState
```
hammeiam, line 150: maybe just throw if `getGaugeData` catches and returns initialstate anyway

src/utils/gauges.ts
```diff
@@ -82,252 +105,263 @@ export const initialGaugesState: Gauges = {
...
+  gaugeRewardCounts: BigNumber[],
+  gaugeMulticallContracts:
+    | MulticallContract<LiquidityGaugeV5>[]
+    | MulticallContract<ChildGauge>[],
+  gaugeRewardsTokens: (string | null)[][],
+  ethCallProvider: MulticallProvider,
+) {
+  return Promise.all(
+    gaugeRewardCounts.map((count, index) =>
+      ethCallProvider.all(
```
hammeiam, line 273: tryall

hammeiam (approved):

src/utils/gauges.ts
```diff
@@ -366,9 +407,345 @@ export async function getGaugeRewardsUserData(
...
+async function getGaugeRewardsFromTokensSidechain(
+  gaugeRewardCounts: BigNumber[],
+  gaugeMulticallContracts: MulticallContract<ChildGauge>[],
+  gaugeRewardsTokens: (string | null)[][],
+  ethCallProvider: MulticallProvider,
+) {
+  return Promise.all(
+    gaugeRewardCounts.map((count, index) =>
+      ethCallProvider.tryAll(
+        enumerate(count.toNumber(), 0).map((num) => {
```
hammeiam, line 739: it seems like this function block won't work as expected. Your `map` should pass a `MulticallCall` to `tryAll`, but is instead passing `{...rewardData, token}`. I believe you'll need to wait for the multicall to complete before structuring the resulting data

src/utils/gauges.ts
```diff
@@ -366,9 +414,347 @@ export async function getGaugeRewardsUserData(
...
+      periodFinish: BN_MSIG_SDL_VEST_END_TIMESTAMP,
+      rate: sdlRate,
+      tokenAddress: SDL_TOKEN_ADDRESSES[chainId].toLowerCase(),
+      isMinter: true,
+    }
+    const gaugeTokens = gaugeRewardTokens[index]
+    const gaugeTokenReward = (gaugeRewards[index] as GaugeTokenRewardData[])
+      .map((reward, tokenIndex) => {
+        if (gaugeTokens[tokenIndex] != null && reward) {
+          const tokenAddress = gaugeTokens[tokenIndex]?.toLowerCase()
```
nguyenaiden, line 544: @penandlim Can I get 2nd look at this line? I'm using the `gaugeTokens` list from the `getGaugeRewardTokens` call on L595. I think this is how the data matches up.

penandlim: `gaugeTokens` here means array of external reward token addresses for this gauge right?

LGTM in that case

src/hooks/useUserGauge.ts
```diff
@@ -95,9 +110,32 @@ export default function useUserGauge(gaugeAddress?: string): UserGauge | null {
...
+            account,
+          )
+          promises.push(childGaugeFactory.mint(gaugeAddress))
+        }
+      } catch (e) {
+        console.error(e)
+        enqueueToast("error", "Unable to claim reward")
       }
 
       return Promise.all(promises)
```
nguyenaiden, line 141: @penandlim, can we eliminate the stale veSDL problem by running some kind of checkpoint function in serial before returning this `Promise.all()`?  Potentially `gaugeContract.user_checkpoint(userAddress)`?

src/hooks/useSidechainGaugeWeightDataOnMainnet.ts
```diff
@@ -0,0 +1,151 @@
...
+  address: string
+  isKilled: boolean
+  gaugeRelativeWeight: BigNumber
+}
+
+export const useSidechainGaugeWeightDataOnMainnet =
+  (): UseQueryResult<SidechainGauges> => {
+    const { library, chainId } = useActiveWeb3React()
+    if (!library || !chainId)
+      throw new Error("Unable to retrieve Sidechain gauge weight data")
```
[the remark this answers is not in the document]
nguyenaiden: Got it. Adding.

penandlim (approved):

nkrulikovsky (approved):

PR #1299, opened 2022-11-03
Replaced BasePool 3 option selection for BasePool Dropdown with entries from Registry

arthur-zhuk:
image
image

* Dropdown of base pools in favor of hardcoded select few pool types
* Adjusted UX logic to adhere to new reqs


src/pages/CreatePool/index.tsx
```diff
@@ -61,9 +81,10 @@ export default function CreatePool(): React.ReactElement {
   const [poolName, setPoolName] = useState<string>("")
   const [poolSymbol, setPoolSymbol] = useState<string>("")
   const [aParameter, setAParameter] = useState<string>("")
-  const [poolType, setPoolType] = useState<PoolType>(PoolType.UsdMeta)
+  const [poolType, setPoolType] = useState<PoolType>(PoolType.Empty)
   const [assetType, setAssetType] = useState<PoolTypes>(2)
```
nkrulikovsky, line 85: I think it would be good to initialize with `PoolTypes.ETH` rather than `2`

src/pages/CreatePool/index.tsx
```diff
@@ -431,107 +436,120 @@ export default function CreatePool(): React.ReactElement {
...
+                              e.target.value
+                            ) {
+                              setTokenInputErrorMsg(
+                                "Duplicate Tokens Not Allowed",
+                              )
+                            } else {
+                              setTokenInputErrorMsg("")
+                            }
+                          }}
+                          onBlur={(e) => void onTokenInputBlur(e.target, index)}
```
nkrulikovsky, line 480: I think it would be good to use `isAddress` to check the valid address on `onChange`event  rather check address on `onBlur` event.

`import { isAddress } from "ethers/lib/utils"`

src/pages/CreatePool/index.tsx
```diff
@@ -361,46 +365,48 @@ export default function CreatePool(): React.ReactElement {
...
-                    )}
-                  <ToggleButton value={PoolType.Base}>
-                    {t("basepool")}
-                  </ToggleButton>
-                </ToggleButtonGroup>
+                    }}
+                  >
+                    <MenuItem value={PoolType.Base}>Base Pool</MenuItem>
+                    {expandedPoolsRemapped.map((pool, i) => (
+                      <MenuItem key={i} value={pool.address}>
```
nkrulikovsky, line 401: Can we use `pool.address` as key rather than `index`?

src/pages/CreatePool/index.tsx
nkrulikovsky: yeah, because we need to fetch only `isAddress` is true.

src/pages/CreatePool/index.tsx
```diff
@@ -361,46 +365,48 @@ export default function CreatePool(): React.ReactElement {
...
+                        const lpTokenAddr = pool.lpToken.address
+                        setPoolType(pool.poolName as PoolType)
+                        setSelectedTokensLength(tokensLength)
+                        setTokenInputs([
+                          ...(Array(maxAmountOfTokens - tokensLength).fill(
+                            "",
+                            0,
+                          ) as string[]),
+                        ])
+                        setMetapoolBasepoolAddr(e.target.value as string)
```
nkrulikovsky, line 394: We are giving poolAddress here. But I think we need to give lp token address in here to display lp token addrss on the last textfield.
Screen Shot 2022-11-03 at 10 42 10 PM

src/pages/CreatePool/index.tsx
```diff
@@ -47,8 +52,23 @@ export type ValidationStatus =
 export default function CreatePool(): React.ReactElement {
   const { t } = useTranslation()
   const theme = useTheme()
-  const poolRegistry = usePoolRegistry()
-  const { account, library, chainId } = useActiveWeb3React()
+  const expandedPools = useContext(ExpandedPoolsContext)
+  const expandedPoolsRemapped = Object.values(expandedPools.data.byName)
+    .map((pool) => ({
```
nguyenaiden, line 57: Could we reverse the order of `.filter` and `.map` and only run `.filter` once? Running `.filter` first can lead to `.map` iterating over less items.

We can combined those filter conditions into a single Boolean. Reduction from O(5n) to O(n) could be noticeable. 

Could be something like this. Up to your discretion to use it verbatim or not.  cc: @hammeiam. Want to see your thoughts on this.

```
const isPoolValid = (pool) => {
  return (
    pool.tokensLength < 4 &&
    pool.isSaddleApproved &&
    !pool.basePoolAddress &&
    !pool.isMigrated &&
    !pool.isPaused
  )
}
const expandedPoolsRemapped = Object.values(expandedPools.data.byName)
  .filter(isPoolValid)
  .map((pool) => ({
    basePoolAddress: pool.basePoolAddress,
    poolName: pool.poolName,
    isSaddleApproved: pool.isSaddleApproved,
    address: pool.poolAddress,
    isPaused: pool.isPaused,
    isMigrated: pool.isMigrated,
    tokensLength: pool.tokens.length,
  }))

File: src/pages/CreatePool/index.tsx [hunk 6]
```diff
@@ -1,39 +1,44 @@
...
 import { useTranslation } from "react-i18next"
 
 export enum PoolType {
   UsdMeta = "usdMetapool",
   BtcMeta = "btcMetapool",
   Base = "basepool",
+  Empty = "",
 }
 
 export type ValidationStatus =
```
nguyenaiden, line 44: I think this should be an enum instead so that on L231, you won't have to cast the type and just use `ValidationStatus.ERROR`. This would also help us use the enum strings instead.

src/pages/CreatePool/index.tsx
```diff
@@ -219,6 +223,15 @@ export default function CreatePool(): React.ReactElement {
       }
       setTokenInfo([...tokenInfo])
     }
+    if (!isAddress(value)) {
+      tokenInfo[index] = {
+        name: "",
+        symbol: "",
+        decimals: 0,
+        checkResult: "error" as ValidationStatus,
```
nguyenaiden, line 231: See comment about `ValidationStatus`.

nguyenaiden (commented):
A few comments.

src/pages/CreatePool/index.tsx
```diff
@@ -353,54 +366,51 @@ export default function CreatePool(): React.ReactElement {
...
+                        setTokenInputs(["", ""])
                       } else {
-                        setTokenInputs([tokenInputs[0]])
+                        const pool =
+                          expandedPools.data.byAddress[e.target.value as string]
+                        if (!pool) {
+                          console.error("Unable to locate pool")
+                          return
+                        }
+                        const tokensLength = pool.tokens.length
```
nguyenaiden, line 388: nit: I think this is more of `tokensCount` rather than `tokensLength`. Up to you!

src/pages/CreatePool/index.tsx
```diff
@@ -353,54 +366,51 @@ export default function CreatePool(): React.ReactElement {
...
-                      </ToggleButton>
-                    )}
-                  <ToggleButton value={PoolType.Base}>
-                    {t("basepool")}
-                  </ToggleButton>
-                </ToggleButtonGroup>
+                    }}
+                  >
+                    <MenuItem value={PoolType.Base}>Base Pool</MenuItem>
+                    {expandedPoolsRemapped.map((pool) => (
```
nguyenaiden, line 404: nit:

```{expandedPoolsRemapped.map(({ address, poolName }) => (
                      <MenuItem key={address} value={address}>
                        {`${poolName} Metapool`}
                      </MenuItem>
                    ))}

nkrulikovsky (approved):

nkrulikovsky (approved):

PR #1305, opened 2022-11-11
update swap state logic with useQuery

nkrulikovsky:
## Describe
This PR replace fetch logic of swap state using `useQuery`. 

Right now, we get swap state when app load and store on redux. If user don't visit pool page, when we don't need to fetch swap state at first. 

`useQuery` stores data to cache, so we don't need to store this data on redux.


src/hooks/useSwapStats.ts
```diff
@@ -0,0 +1,89 @@
...
+    async () => {
+      const res = await fetch(`${swapStatsURI}`)
+      if (res.status >= 200 && res.status < 300) {
+        return res.json()
+      }
+      throw new Error("Unable to fetch swap stats from IPFS")
+    },
+    { retry: 3, refetchInterval: BLOCK_TIME * 3 },
+  )
+  if (isSuccess) {
```
ug02fast, line 46: instead of manually handling undefined you should be able to use `selector` as described here: https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option

this only runs if query isn't undefined

ug02fast (approved):
looks good

src/hooks/useSwapStats.ts
nkrulikovsky: I didn't know about the `select`. Will update!

ug02fast: I've never used it myself so it's up to you

nkrulikovsky: I love this way.

src/hooks/useSwapStats.ts
```diff
@@ -0,0 +1,85 @@
...
+    ["swapStats"],
+    async () => {
+      const res = await fetch(`${swapStatsURI}`)
+      if (res.status >= 200 && res.status < 300) {
+        return res.json()
+      }
+      throw new Error("Unable to fetch swap stats from IPFS")
+    },
+    {
+      select: useCallback(
```
hammeiam, line 39: Why is this a `useCallback` instead of a plain function?

hammeiam (changes requested):
I don't see the request being made in the fleek build, is something missing?

Screen Shot 2022-11-15 at 10 10 56 PM

nkrulikovsky:
@hammeiam 
I believe we are fetching `swap state` on `PoolInfoCard`, so when we open `withdraw` or `deposit` page, fetching will be occured.

src/hooks/useSwapStats.ts
nkrulikovsky: https://github.com/TkDodo/blog-comments/discussions/57#discussioncomment-3759612
> we do memoize the result of select, but only if:
> 
> - data hasn't changed
> 
> - the functional identity hasn't changed
> 
> the first one should be obvious - select needs to re-run if we get new data from the server.
> the second one is there to ensure that if you closure over dependencies, you also get a correct result. For example, if you select an item from a list, and the id of that item comes in as a prop, we need to re-run select when that prop changes. That's where useCallback helps, as I've shown in the example: The functional identity will change only when id changes, so we know when to re-run select.

hammeiam:
> @hammeiam I believe we are fetching `swap state` on `PoolInfoCard`, so when we open `withdraw` or `deposit` page, fetching will be occurred.

`swapStats` is used in `PoolInfoCard` as well as in `usePoolData`, where it is used for `PoolOverview` to display APY and Volume. It would be good to remove `swapStats` from `usePoolData` eventually, but for now your new hook needs to make `swapStats` available to `usePoolData` as well to avoid regressions.

src/hooks/useSwapStats.ts
```diff
@@ -0,0 +1,85 @@
...
+        (data: SwapStatsReponse) => formatSwapStats(data),
+        [],
+      ),
+      retry: 3,
+      refetchInterval: BLOCK_TIME * 3,
+    },
```
hammeiam, line 45: I think you also need to add arguments for either `staleTime` or `cacheTime`, since the call is made every time that the hook is called. Verify this yourself by navigating to one pool's deposit page, then to another pool's deposit page. You will see two requests for the same data.

nkrulikovsky: I think some of unexpected re-fetch is occurring from `refetchOnWindowFocus`. I disable this feature.

hammeiam (approved):
Looks great! Thanks for your changes

PR #1306, opened 2022-11-13
Bug fix for withdraw amount

nkrulikovsky:


src/hooks/usePoolData.ts
```diff
@@ -248,14 +253,25 @@ export default function usePoolData(name?: string): PoolDataHookReturnType {
...
-        }))
+        const userPoolTokens =
+          expandedPool.tokens.map((token, i) => ({
+            isOnTokenLists: token.isOnTokenLists,
+            symbol: token.symbol,
+            name: token.name,
+            decimals: token.decimals,
+            value: userPoolTokenBalances[i],
+            address: token.address,
+          })) || []
```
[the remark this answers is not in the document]
nkrulikovsky: Because ` expandedPool.underlyingTokens` is optional, I gave empty array rather than returning undefined.
But actually, expandedPool is always returning value so after clearing null value, we can remove this empty array.

src/hooks/usePoolData.ts
```diff
@@ -157,7 +168,10 @@ export default function usePoolData(name?: string): PoolDataHookReturnType {
           ...emptyPoolData,
           name: poolName || "",
         })
-        setUserShare(null)
+        setUserShare((prev) => ({
+          ...prev,
+          name: poolName || "",
```
ug02fast, line 173: why set `poolName` here if this block is for when `poolName == null`? Why is it necessary to explicitly set name here?

ug02fast (approved):

src/hooks/usePoolData.ts
nkrulikovsky: You are correctly, I followed only the poolData logic but we don't need to give poolName here because we initialized poolName with "".

hammeiam (approved):

PR #1308, opened 2022-11-16
Protec ChildGauge contract call against certain chains.

ug02fast:
Only execute this call when on Arbitrum or Optimism. Currently causing bug that prevents pages from rendering.

### DIFF
--- a/src/utils/gauges.ts
+++ b/src/utils/gauges.ts
@@ -142,6 +142,7 @@ export async function getGaugeData(
         account,
       )
     } else {
+      console.error(`Gauges are not supported on chainId: ${chainId}`)
       return initialGaugesState
     }
   } catch (e) {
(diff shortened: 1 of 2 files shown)


src/utils/gauges.ts
```diff
@@ -130,14 +133,17 @@ export async function getGaugeData(
...
+    } else if (shouldLoadChildGauges(chainId)) {
       return buildGaugeDataSidechain(
         library,
         chainId,
         basicPools,
         registryAddresses,
         account,
       )
+    } else {
+      console.error(`Gauges are not supported on chainId: ${chainId}`)
```
nguyenaiden: I think this is better.  `console.error("Gauges are not supported on ChainId[chainId].")`

nguyenaiden (approved):
+1 upon update.

PR #1317, opened 2022-12-20
Use '_write' fn for gauges data, update controller abi to view, format gauge names

hammeiam:
Changes to support correct operation of sidechain gauges:
- Use '_write' fn when fetching gauges data
- Update controller abi to view
- Format gauge names to include chain name (for sidechains)
- Include sidechain root gauges in voting options

Screen Shot 2022-12-19 at 6 15 10 PM
Screen Shot 2022-12-19 at 6 15 04 PM


src/pages/VeSDL/OnChainVote.tsx
```diff
@@ -77,8 +80,14 @@ export default function OnChainVote({
             gaugeName: gaugeName,
           }
       })
+    ;(sidechainGaugesInfo?.gauges || []).forEach(({ address, displayName }) => {
```
hammeiam, line 83: @penandlim here's that semicolon thing

penandlim (approved):

PR #1334, opened 2023-02-06
Fix gauge claiming

hammeiam:
Adjust logic in the `TokenClaimDialog` to simply reuse the claim function from the `useUserGauge` hook.


mjdatadog (approved):
Claim successful

penandlim (approved):

PR #1344, opened 2023-03-07
Added base testnet, turned on permissionless pools for aurora

mjdatadog:


hammeiam (approved):
Approved, but you mentioned that users won't be able to use metamask? Should we only activate this for WalletConnect then?

PR #1366, opened 2023-05-05
Rainbow integration

nkrulikovsky:
_**Description**_

- Install `rainbowkit` and `wagmi` deps.
- create `config` file for `wagmiClient`

### DIFF
--- a/src/connectors/config.ts
+++ b/src/connectors/config.ts
@@ -1,0 +1,115 @@
+import { STALL_TIMEOUT, SUPPORTED_NETWORKS } from "../constants/networks"
+import {
+  arbitrum,
+  aurora,
+  avalanche,
+  evmos,
+  evmosTestnet,
+  fantom,
+  goerli,
+  hardhat,
+  mainnet,
+  optimism,
+  optimismGoerli,
+} from "wagmi/chains"
+import {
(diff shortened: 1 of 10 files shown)


src/connectors/config.ts
```diff
@@ -0,0 +1,115 @@
...
+  throw new Error(
+    "You must define REACT_APP_ALCHEMY_API_KEY in your environment",
+  )
+}
+
+const { chains, provider } = configureChains(Object.values(chain), [
+  alchemyProvider({
+    apiKey: alchemyKey,
+    stallTimeout: STALL_TIMEOUT,
+    priority: process.env.REACT_APP_PROVIDER_ID === "ALCHEMY" ? 0 : 2,
```
hammeiam, line 80: When would the alchemy provider not be our top priority? Especially since we throw an error above if an alchemy key isn't set.

hammeiam (approved):
Looks great! Now for the hard part

src/connectors/config.ts
nkrulikovsky: You are correct. We shouldn't throw the error when there is no alchemyKey so that we can use jsonProvider if we missing the alchemyKey

PR #1370, opened 2023-05-13
Wagmi provider

nkrulikovsky:


src/pages/CoinbasePayTest.tsx
hammeiam, line 1: If we're removing coinbase pay (or just the custom button?) then we should also fully remove the test as well

hammeiam (approved):

PR #1371, opened 2023-05-14
Update walletconnect button

nkrulikovsky:


package.json
```diff
@@ -55,7 +55,8 @@
     "react-toastify": "^8.1.1",
     "redux": "^4.1.0",
     "redux-localstorage-simple": "^2.4.0",
-    "wagmi": "^0.12.12",
+    "viem": "^0.3.22",
+    "wagmi": "^1.0.2",
```
hammeiam, line 59: I don't think you want to upgrade to wagmi or rainbowkit 1.0. The smart contract responses from both changes from BigNumber to BigInt, so you would need to change all numbers everywhere in the app. Unless it's absolutely necessary, I think we should only use <1.0 for both.

nkrulikovsky: You are correct. Because of big change `ether.js v6` wagmi and rainbowkit 1.0 is returning `bigint`. I upgrade these libraries because the webpack issues we discussed (https://github.com/wagmi-dev/wagmi/issues/2300). 

Now, ehters version is `6.3.0` and `typechain` released its [version for `6.0`](https://github.com/dethcrypto/TypeChain/releases/tag/%40typechain%2Fethers-v6%400.3.2) which we can install by `@typechain/ethers-v6`.

Also in most of cases, we are converting `BigNumber` to `number` using mostly `formatBNToString`, `formatBNToPercentString`, `calculatePrice`   across over all our codebase. which can migrate easily by changing `BigInt` or [`BigNumberish`](https://docs.ethers.org/v5/single-page/#/v5/api/utils/bignumber/-%23-BigNumber--types).
So, I think migration won't take much time.

hammeiam (approved):
Code looks good to me. You didn't have to change anything in package.json to fix the webpack issue?

hammeiam:
Looks like lint and build are both failing, could you resolve those issues?

nkrulikovsky:
Yes, we don't need to update wagmi to `1.0.0`. Yes, I'll fix the lint issue.

